PR #1, opened 2022-08-03
DIS-1334 Fix default array constructors for dynamic or null terminated arrays

Horofic:


Schamper (approved):

Schamper:
Maybe just rename the title/squash commit to "Fix default array constructors for dynamic or null terminated arrays"

PR #10, opened 2022-10-07
Update license acc. license file

fabaff:


Schamper:
Thanks for pointing this out, looks like we missed this! Looks like we forgot to change out this `setup.py` from the old dissect.cstruct repo to the newer `setup.py` template. I've created a new MR to fix the overarching issue at #11, so I'll close this one.

PR #11, opened 2022-10-17
Replace old setup.py

Schamper:
(DIS-1610)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,22 +1,6 @@
 from setuptools import setup
 
-with open("README.md") as fh:
-    long_description = fh.read()
-
 setup(
     name="dissect.cstruct",
-    author="Fox-IT",
-    description="Structure parsing in Python made easy.",
-    long_description=long_description,
-    long_description_content_type="text/markdown",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -1,22 +1,6 @@
...
 setup(
     name="dissect.cstruct",
-    author="Fox-IT",
-    description="Structure parsing in Python made easy.",
-    long_description=long_description,
-    long_description_content_type="text/markdown",
-    license="MIT",
-    keywords="cstruct struct dissect structure binary pack packer unpack unpacker parser parsing",
-    url="https://github.com/fox-it/dissect.cstruct",
     packages=["dissect.cstruct", "dissect.cstruct.types"],
```
Miauwkeru: Is there a reason why this project specifically does not use:
```suggestion
    packages=list(map(lambda v: "dissect." + v, find_packages("dissect"))),
```
?

Miauwkeru (changes requested):

Miauwkeru (approved):
LGTM

PR #13, opened 2022-11-01
Configure the minimum python version

pyrco:
The Python version dissect is developed and tested against is currently 3.9. It
is set as minimum version and not also as maximum version because newer
versions (like 3.10) might work and because it makes transitioning to newer
Python versions easier.

(DIS-1490)


Miauwkeru (approved):
LGTM

PR #18, opened 2022-11-21
Check if field is pointer before checking if field is an array

MrYoranimo:
Considering the following field definition:

```c
char *text[AUDIT_MAX_ARGS];
```

Previously, `TokenParser._parse_field` would first check if a field is an array, and then if a field is a pointer. This would result in the field being interpreted as a pointer to an array. However, this should have been interpreted as an array of pointers.

This issue became prevalent in `Compiler.gen_read_block`, that would first check if a field's type is Pointer and only then if the field's type is an Array of Pointers. As far as I can tell, this condition cannot be reached with the current implementation of `_parse_field`.

This change performs the check if the field is a pointer before the check if the field is an array. This would result in the field being interpreted as an Array of Pointers.

### DIFF
--- a/dissect/cstruct/compiler.py
+++ b/dissect/cstruct/compiler.py
@@ -51,7 +51,7 @@ class {name}(Structure):
         self.source = source
         super().__init__(cstruct, structure.name, structure.fields, anonymous=structure.anonymous)
 
-    def _read(self, stream):
+    def _read(self, stream, context=None):
         r = OrderedDict()
         sizes = {{}}
         bitreader = BitBuffer(stream, self.cstruct.endian)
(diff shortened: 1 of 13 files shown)


jasper-vdhoven:
Thanks for finding and resolving this whilst I was trying to write a shitty parser :heart:

Schamper:
Nice find! Would you mind adding a unit test too? I think `tests/test_pointer.py` is probably the best place for it.

MrYoranimo:
Thanks @Schamper! I was unable to create a working unit test without making some other changes to the code (and potentially fixing some more bugs). Could you take a look at the new changes if you can find some time?

Schamper:
Ah yes, this touches some changes I've been meaning to make. I'll review them shortly!

MrYoranimo:
@DissectBot agree

Schamper:
@MrYoranimo I'm letting one of my colleagues do a final review and then we should be good to go!

Miauwkeru (approved):
LGTM!

PR #20, opened 2022-11-25
Fix bitbuffer to support `char`

sud0woodo:
(DIS-1674)

Added check to see if a field is a `char` type and convert the field to int so we can perform bitwise operations.

Also changed the check for the little vs big endianness as it was first checking if it wasn't big endian, improves reading the code.

### DIFF
--- a/tests/test_bitfield.py
+++ b/tests/test_bitfield.py
@@ -246,3 +246,29 @@ def test_bitfield_with_enum_or_flag(compiled):
     assert obj.d == cs.Flag8.A | cs.Flag8.B
 
     assert obj.dumps() == buf
+
+
+def test_bitfield_char(compiled):
+    cdef = """
+    struct test {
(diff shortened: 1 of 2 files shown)


tests/test_bitfield.py
```diff
@@ -253,22 +253,22 @@ def test_bitfield_char(compiled):
...
         char    d[4];
     };
     """
 
     cs = cstruct.cstruct()
-    cs.load(cdef)
+    cs.load(cdef, compiled=compiled)
 
-    assert verify_compiled(cs.test, compiled)
+    assert verify_compiled(cs.test, compiled=compiled)
```
Schamper:
```suggestion
    assert verify_compiled(cs.test, compiled)
```
Just for consistency sake.

Schamper (changes requested):

Schamper (approved):

Issue #22, opened 2023-01-28
Array sizes are not enforced on writing the structure instance


skarndev:
Is there any particular reason there is no validation performed on the array sizes for static structures, where array size is also static?

It sort of produces the mismatch between what is actually written and the type schema.

Schamper:
No particular reason, most likely just an oversight and/or never was an issue in our current use cases. However it does sound like a nice improvement! Feel free to submit a PR for this, otherwise we'll see if we can add this in ourselves soon.

I actually believe there are quite a few quality of life improvements that can be made in our writing functionality, but they often get less attention since our primary use case for cstruct is parsing. The few times we use it for writing, we tend to work around some of these issues in the surrounding code and we forget to fix it in cstruct itself. I.e. sane default values for empty fields.

If you run into any more problems or possible improvements, please do create more GitHub issues!

PR #23, opened 2023-01-28
Raise an exception when the static size of an array field does not match the size of the provided list instead of silently writing incorrectly sized data.

skarndev:
Addresses issue #22.

### DIFF
--- a/dissect/cstruct/exceptions.py
+++ b/dissect/cstruct/exceptions.py
@@ -14,6 +14,5 @@ class NullPointerDereference(Error):
     pass
 
 
-class ArraySizeMismatch(Error):
+class ArraySizeError(Error):
     pass
-
--- a/dissect/cstruct/exceptions.py
+++ b/dissect/cstruct/exceptions.py
@@ -12,3 +12,8 @@ class ResolveError(Error):
 
 class NullPointerDereference(Error):
     pass
+
+
+class ArraySizeMismatch(Error):
+    pass
+
--- a/dissect/cstruct/types/base.py
+++ b/dissect/cstruct/types/base.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from io import BytesIO
 from typing import Any, BinaryIO, List, TYPE_CHECKING
 
-from dissect.cstruct.exceptions import ArraySizeMismatch
+from dissect.cstruct.exceptions import ArraySizeError
 from dissect.cstruct.expression import Expression
 
 if TYPE_CHECKING:
@@ -52,7 +52,7 @@ class BaseType:
             The resulting bytes.
 
         Raises:
-            ArraySizeMismatch: Raised when len(data) does not match the size of a statically sized array field.
+            ArraySizeError: Raised when ``len(data)`` does not match the size of a statically sized array field.
         """
         out = BytesIO()
         self._write(out, data)
@@ -84,7 +84,7 @@ class BaseType:
             The amount of bytes written.
 
         Raises:
-            ArraySizeMismatch: Raised when len(data) does not match the size of a statically sized array field.
+            ArraySizeError: Raised when ``len(data)`` does not match the size of a statically sized array field.
         """
         return self._write(stream, data)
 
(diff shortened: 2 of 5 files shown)


dissect/cstruct/exceptions.py
```diff
@@ -12,3 +12,8 @@ class ResolveError(Error):
 
 class NullPointerDereference(Error):
     pass
+
+
+class ArraySizeMismatch(Error):
```
Schamper:
```suggestion
class ArraySizeError(Error):
```
I've been trying to think if there was some Python builtin error we could use but came up blank. I think it's better to rename this to be more consistent with other error names though.

dissect/cstruct/types/base.py
```diff
@@ -3,6 +3,7 @@
 from io import BytesIO
 from typing import Any, BinaryIO, List, TYPE_CHECKING
 
+from dissect.cstruct.exceptions import ArraySizeMismatch
```
Schamper:
```suggestion
from dissect.cstruct.exceptions import ArraySizeError
```

dissect/cstruct/types/base.py
```diff
@@ -49,6 +50,9 @@ def dumps(self, data: Any) -> bytes:
 
         Returns:
             The resulting bytes.
+
+        Raises:
+            ArraySizeMismatch: Raised when len(data) does not match the size of a statically sized array field.
```
Schamper:
```suggestion
            ArraySizeError: Raised when ``len(data)`` does not match the size of a statically sized array field.
```

dissect/cstruct/types/base.py
```diff
@@ -78,6 +82,9 @@ def write(self, stream: BinaryIO, data: Any) -> int:
 
         Returns:
             The amount of bytes written.
+
+        Raises:
+            ArraySizeMismatch: Raised when len(data) does not match the size of a statically sized array field.
```
Schamper:
```suggestion
            ArraySizeError: Raised when ``len(data)`` does not match the size of a statically sized array field.
```

dissect/cstruct/types/base.py
```diff
@@ -158,6 +165,9 @@ def _write(self, stream: BinaryIO, data: List[Any]) -> int:
         if self.null_terminated:
             return self.type._write_0(stream, data)
 
+        if not self.dynamic and self.count != (arr_size := len(data)):
+            raise ArraySizeMismatch(f"Expected static array size {self.count}, got {arr_size} instead.")
```
Schamper:
```suggestion
            raise ArraySizeError(f"Expected static array size {self.count}, got {actual_size} instead.")
```

dissect/cstruct/types/base.py
```diff
@@ -158,6 +165,9 @@ def _write(self, stream: BinaryIO, data: List[Any]) -> int:
         if self.null_terminated:
             return self.type._write_0(stream, data)
 
+        if not self.dynamic and self.count != (arr_size := len(data)):
```
Schamper:
```suggestion
        if not self.dynamic and self.count != (actual_size := len(data)):
```

dissect/cstruct/exceptions.py
```diff
@@ -12,3 +12,8 @@ class ResolveError(Error):
...
+    pass
+
```
Schamper:
```suggestion
```
Make the linter happy.

tests/test_basic.py
```diff
@@ -417,3 +417,20 @@ def test_dunder_bytes():
...
+    """
+    c = cstruct.cstruct(endian=">")
+    c.load(d)
+
+    a = c.test(a=[1, 2, 3])
+
+    with pytest.raises(ArraySizeMismatch):
+        a.dumps()
+
+
```
Schamper:
```suggestion
```
Make the linter happy.

tests/test_basic.py
```diff
@@ -417,3 +417,20 @@ def test_dunder_bytes():
...
+def test_report_array_size_mismatch():
+    d = """
+    struct test {
+        uint8   a[2];
+    };
+    """
+    c = cstruct.cstruct(endian=">")
+    c.load(d)
+
+    a = c.test(a=[1, 2, 3])
```
Schamper:
```suggestion
    cdef = """
    struct test {
        uint8   a[2];
    };
    """
    cs = cstruct.cstruct(endian=">")
    cs.load(cdef)

    a = cs.test(a=[1, 2, 3])
```

Schamper (changes requested):

Schamper:
@skarnproject thanks, if you could give an ack on the CLA this is good to go!

skarndev:
@DissectBot agree

Schamper (approved):

PR #24, opened 2023-01-30
Add support for two-dimensional arrays

MrYoranimo:
A field may be defined as a two-dimensional array (array of arrays) in case of an array of strings, which are actually character arrays.

The proposed changes allow the parser to read field names with a secondary set of square brackets with an optional length. Furthermore, the compiler is updated in order to actually read the child arrays.


Schamper:
Thanks! Please review a suggestion I just made on your branch to allow arbitrarily sized multi-dimensional arrays instead of just two-dimensional.

MrYoranimo:
That feels like a better solution indeed, LGTM!

Schamper (approved):

PR #27, opened 2023-02-25
Pass context when reading a compiled Structure or Array

MrYoranimo:
After allowing nested arrays in previous commits, only the parent type would have access to the context while resolving expressions in order to calculate their size.

This changeset updates the compiler so that other fields' values are passed as context so that sublevel arrays can also resolve their size from an expression correctly. A test has been written to verify this change.


Schamper (approved):
Good catch!

PR #29, opened 2023-04-02
Make the Instance class dict compatible to allow for JSON serialization.

### LINKED ISSUE #98: Make the Instance class dict compatible to allow for JSON serialization. PR #29
This issue is to follow up on PR #29

qkaiser:
I have a specific use case where I want to serialize dissect.cstruct instances to JSON.

This MR introduces two changes to make both `Instance` and `EnumInstance` inherit from `dict`, allowing users of the library to do things like:

```python
my_struct = cstruct()
my_struct.load(definition)
record = my_struct.Record(data)
print(json.dumps(record, cls=CustomEncoder))
```

Since these instances can contain `bytes` attributes and that JSON does not support bytes, the specificities of the JSON encoding is left to the user of `dissect.cstruct`. In the attached example, a UTF-8 decoding with surrogate escape is used but we could also imagine a base64 encoding if the structure holds lots of raw binary data.

A demo example is provided in `examples/mirai_json.py`

These changes do not introduce API changes nor do they break the test suite. However, I'm open to writing unit tests if you're open to merge this MR :)

Thanks again for this wonderful project !

### DIFF
--- a/dissect/cstruct/types/enum.py
+++ b/dissect/cstruct/types/enum.py
@@ -84,12 +84,14 @@ class Enum(RawType):
         return [self.default() for _ in range(count)]
 
 
-class EnumInstance:
+class EnumInstance(dict):
     """Implements a value instance of an Enum"""
 
     def __init__(self, enum: Enum, value: int):
         self.enum = enum
         self.value = value
+        kwargs = {self.enum.name : self.value}
+        dict.__init__(self, **kwargs)
 
     def __eq__(self, value: Union[int, EnumInstance]) -> bool:
         if isinstance(value, EnumInstance) and value.enum is not self.enum:
--- a/dissect/cstruct/types/instance.py
+++ b/dissect/cstruct/types/instance.py
@@ -4,12 +4,13 @@ from typing import Any, BinaryIO, Dict
 from dissect.cstruct.types import BaseType
 
 
-class Instance:
+class Instance(dict):
     """Holds parsed structure data."""
 
     __slots__ = ("_type", "_values", "_sizes")
 
     def __init__(self, type_: BaseType, values: Dict[str, Any], sizes: Dict[str, int] = None):
+        dict.__init__(self, **values)
         # Done in this manner to check if the attr is in the lookup
         object.__setattr__(self, "_type", type_)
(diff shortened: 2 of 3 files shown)


qkaiser:
Just fixed my code with your linter config.

pyrco:
Thanks for the nice idea! Unfortunately there are some issues with subclassing `Instance` from `dict` as we have plans to merge the `Instance` and  `Structure` classes into a single class. Having that new class be based off of `dict` would be difficult.

Would it be feasible to have an `as_dict()` function instead? The EnumInstance could then be dealt with in a `CustomEncoder` as I can imagine other users wanting to have just the value of an enum (possibily even just the name) instead of a key/value pair.

pyrco (changes requested):
Thanks for the nice idea! Unfortunately there are some issues with subclassing `Instance` from `dict` as we have plans to merge the `Instance` and `Structure` classes into a single class. Having that new class be based off of `dict` would be difficult.

Would it be feasible to have an `as_dict()` function instead? The `EnumInstance` could then be dealt with in a `CustomEncoder` as I can imagine other users wanting to have just the value of an enum (possibly even just the name) instead of a key/value pair.

qkaiser:
> Thanks for the nice idea! Unfortunately there are some issues with subclassing `Instance` from `dict` as we have plans to merge the `Instance` and `Structure` classes into a single class. Having that new class be based off of `dict` would be difficult.
> 
> Would it be feasible to have an `as_dict()` function instead? The `EnumInstance` could then be dealt with in a `CustomEncoder` as I can imagine other users wanting to have just the value of an enum (possibly even just the name) instead of a key/value pair.

I see. We'll probably implement the serialization on our end then, np :)

Since you plan on merging `Instance` and `Structure`, I just wanted to mention that outside of JSON serialization, it can also be a bottleneck for users of  `dissect.cstruct` wanting to transfer `Instance` objects between processes since multiprocessing in Python rely on pickle.

I'll close this once we're sure of the direction to take.

Again, thanks for all your work on this project !

Miauwkeru:
@qkaiser is this still an issue for you with the new dissect cstruct v4?

PR #36, opened 2023-06-07
make enums support bitwise operations

cecinestpasunepipe:


cecinestpasunepipe:
Never mind, there is already a Flag type.

Issue #32, opened 2023-04-13
Support integer suffixes


d3dave:
Integers can have unsigned (`U`/`u`), long (`L`/`l`), and long-long (`LL`/`ll`) suffixes. When attempting to parse these we get a `ValueError` (in `expression.py`):

```console
ValueError: invalid literal for int() with base 10: '1U'
```

sMezaOrellana:
I think https://github.com/fox-it/dissect.cstruct/pull/37 closes this

Schamper:
I believe so too, I had hoped GitHub would automatically close this.

Schamper:
I believe so too, I had hoped GitHub would automatically close this.

PR #37, opened 2023-06-12
Expression parser support for order of operations and unary negation operator

sMezaOrellana:
I realized that the expression parser class was not able to parse more 'complex' expressions.
Things like  "4 * 1 + 1" would not equal "5".  The previous expression parser did not support the order of operations.
I implemented this using a shunting yard parsing algorithm[1] for evaluating arithmetic expressions.  I also added support for unary operators like '-' and '~'.
This implementation mostly follows the order of precedence defined in the C standard[2]. I added some more tests.

Fan of this project.

[1] 
https://en.wikipedia.org/wiki/Shunting_yard_algorithm
[2]
https://en.cppreference.com/w/c/language/operator_precedence


Schamper:
Thanks for the amazing PR! I'm unfortunately personally out of office for another week but I've assigned @pyrco to review it. You should also receive a contributor agreement message shortly as a reply on this PR (our bot is unfortunately still a human and out of office too ;))

One quick question I already have: is it possible to support integer suffixes with your parser (#32)?

sMezaOrellana:
@DissectBot agree []

sMezaOrellana:
> One quick question I already have: is it possible to support integer suffixes with your parser (#32)?

Yes this could be supported by slightly modifying the tokenizer. I will look at it

sMezaOrellana:
I have added support for suffixes. This means that when we encounter a number like '10ull' this gets parsed into '10'.
Suffixes are parsed according to [1] This means a number can have the following suffixes  `u, l, ll, ul, ull, lu, llu` either in uppercase or lowercase. There are some edge cases to consider. If we look at the specifications [2]:
```
 std::cout << -9223372036854775808u << '\n'; // unary minus applied to unsigned
                // value subtracts it from 2^64, this evaluates as 9223372036854775808
```
The upper code would output `9223372036854775808` however because of how Python works we output:
` -9223372036854775808`
[1]
https://www.ibm.com/docs/de/zos/2.1.0?topic=literals-integer

[2]
https://en.cppreference.com/w/cpp/language/integer_literal

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+from dissect.cstruct.exceptions import ExpressionParserError
 
 if TYPE_CHECKING:
     from dissect.cstruct import cstruct
 
 
+class ExpressionTokenizer:
+    def __init__(self, expression: str):
+        self.expression = expression
+        self.i: int = 0
```
Schamper:
```suggestion
        self.i = 0
```
I think this type hint is redundant, also perhaps consider a more descriptive name for `i`, i.e. `pos` or `idx`.

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+        self.tokens = []
+
+    def equal(self, token: str, expected: str) -> bool:
```
Schamper:
```suggestion
    def equal(self, token: str, expected: Union[str, set[str]]) -> bool:
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+        return token.isalnum()
+
+    def alpha(self, token: str) -> bool:
+        return token.isalpha()
+
+    def digit(self, token: str) -> bool:
+        return token.isdigit()
+
+    def digitHex(self, token: str) -> bool:
+        return token.isdigit() or token in {"a", "b", "c", "d", "e", "f", "A", "B", "C", "D", "E", "F"}
```
Schamper:
```suggestion
    def hexdigit(self, token: str) -> bool:
        return token in string.hexdigits
```
And add a `import string` at the top.

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+
+    def operator(self, token) -> bool:
+        op: set[str] = {"*", "/", "+", "-", "%", "&", "^", "|", "(", ")", "~"}
+        return token in op
```
Schamper:
```suggestion
        return token in {"*", "/", "+", "-", "%", "&", "^", "|", "(", ")", "~"}
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+
+    def match(self, func: Callable = None, expected: str = None, consume: bool = True, append: bool = True) -> bool:
+        if self.outOfBounds():
+            return False
+
+        token: str = self.getToken()
```
Schamper:
```suggestion
        token = self.get_token()
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+        return token.isdigit()
+
+    def digitHex(self, token: str) -> bool:
+        return token.isdigit() or token in {"a", "b", "c", "d", "e", "f", "A", "B", "C", "D", "E", "F"}
+
+    def operator(self, token) -> bool:
+        op: set[str] = {"*", "/", "+", "-", "%", "&", "^", "|", "(", ")", "~"}
+        return token in op
+
+    def match(self, func: Callable = None, expected: str = None, consume: bool = True, append: bool = True) -> bool:
```
Schamper:
```suggestion
    def match(self, func: Optional[Callable[[str], bool] = None, expected: Optional[str] = None, consume: bool = True, append: bool = True) -> bool:
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+    def alpha(self, token: str) -> bool:
+        return token.isalpha()
+
+    def digit(self, token: str) -> bool:
+        return token.isdigit()
+
+    def digitHex(self, token: str) -> bool:
+        return token.isdigit() or token in {"a", "b", "c", "d", "e", "f", "A", "B", "C", "D", "E", "F"}
+
+    def operator(self, token) -> bool:
```
Schamper:
```suggestion
    def operator(self, token: str) -> bool:
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+        if func and func(token):
+            if append:
+                self.tokens.append(token)
+            if consume:
+                self.consume()
+            return True
+
+        return False
+
+    def consume(self):
```
Schamper:
```suggestion
    def consume(self) -> None:
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+        self.i += 1
+
+    def outOfBounds(self):
```
Schamper:
```suggestion
    def out_of_bounds(self) -> bool:
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+        return self.i >= len(self.expression)
+
+    def getToken(self) -> str:
```
Schamper:
```suggestion
    def get_token(self) -> str:
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+            raise Exception(f"Out of bounds index: {self.i}, length: {len(self.expression)}")
+        return self.expression[self.i]
+
+    def tokenize(self) -> list[str]:
+        token = ""
+
+        # Loop over expression runs in linear time
+        while not self.outOfBounds():
+            # if token is a single character operand add it to tokens
+            if self.match(func=self.operator):
```
Schamper:
```suggestion
            if self.match(self.operator):
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+                    self.match(expected={"l", "L"}, append=False)
+
+                elif self.match(expected={"l", "L"}, append=False):
+                    self.match(expected={"l", "L"}, append=False)
+                    self.match(expected={"u", "U"}, append=False)
+                else:
+                    pass
+
+                # number cannot end on x or b in the case of binary or hexadecimal notation
+                assert token[-1] != "x" and token[-1] != "b" and token[-1] != "X"
```
Schamper: Can you change this into an exception? You could use the `ExpressionTokenizerError` you added.

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+
+    def consume(self):
+        self.i += 1
+
+    def outOfBounds(self):
+        return self.i >= len(self.expression)
+
+    def getToken(self) -> str:
+        if self.i >= len(self.expression):
+            raise Exception(f"Out of bounds index: {self.i}, length: {len(self.expression)}")
```
Schamper:
```suggestion
            raise ExpressionTokenizerError(f"Out of bounds index: {self.i}, length: {len(self.expression)}")
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+                token = ""
+            # if token is length 2 operand make sure next character is part of length 2 operand append to tokens
+            elif self.match(expected=">", append=False) and self.match(expected=">", append=False):
+                self.tokens.append(">>")
+            elif self.match(expected="<", append=False) and self.match(expected="<", append=False):
+                self.tokens.append("<<")
+            elif self.match(expected=" ", append=False):
+                continue
+            else:
+                raise Exception(f"Tokenizer does not recognize following token '{self.expression[self.i]}'")
```
Schamper:
```suggestion
                raise ExpressionTokenizerError(f"Tokenizer does not recognize token '{self.expression[self.i]}'")
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+            "+": 4,
+            "-": 4,
+            "*": 5,
+            "/": 5,
+            "%": 5,
+            "u": 6,
+            "~": 6,
+            "sizeof": 6,
+        }  # Operator precedence levels
+        return p[o1] >= p[o2]
```
Schamper:
```suggestion
    precedence_levels = {
        "^": 1,
        "&": 5,
        "<<": 3,
        ">>": 3,
        "+": 4,
        "-": 4,
        "*": 5,
        "/": 5,
        "%": 5,
        "u": 6,
        "~": 6,
        "sizeof": 6,
    }

    def precedence(self, o1: str, o2: str) -> bool:
        return self.precedence_levels[o1] >= self.precedence_levels[o2]
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
-        context = context or {}
-        levels = []
-        buf = ""
-
-        for i in range(len(self.expression)):
-            if self.expression[i] == "(":
-                levels.append(buf)
-                buf = ""
-                continue
+    def evaluate_exp(self, context: Dict[str, int] = None) -> None:
```
Schamper:
```suggestion
    def evaluate_exp(self, context: Optional[dict[str, int]] = None) -> None:
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+            right = self.queue.pop(-1)
+            res = ~right
+        else:
+            right = self.queue.pop(-1)
+            left = self.queue.pop(-1)
+            res = self.operators[operator](left, right)
+        self.queue.append(res)
 
-            buf += self.expression[i]
+    def isNumber(self, currentToken: str) -> bool:
```
Schamper:
```suggestion
    def is_number(self, current_token: str) -> bool:
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+        return currentToken.isnumeric() or (
+            len(currentToken) > 2
+            and currentToken[0] == "0"
+            and (currentToken[1] == "x" or currentToken[1] == "X" or currentToken[1] == "b" or currentToken[1] == "o")
```
Schamper:
```suggestion
            and currentToken[1] in ("x", "X", "b", "B", "o", "O")
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+        )
 
-        return self.evaluate_part(buf, context)
+    def parseErr(self, condition: bool, error: str) -> None:
+        if condition:
+            raise ExpressionParserError(error)
 
-    def evaluate_part(self, buf: str, context: Dict[str, int]) -> int:
-        buf = buf.strip()
+    def evaluate(self, context: Dict[str, int] = None) -> int:
```
Schamper:
```suggestion
    def evaluate(self, context: Optional[dict[str, int]] = None) -> int:
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
-        # they must be wrapped in brackets, e.g.: 2 * (-5).
-        #
-        # To have full support for the negation operator a proper expression
-        # parser must be build.
-        if buf.startswith("-") and buf[1:].isnumeric():
-            return int(buf)
+        self.stack = []
+        self.queue = []
+        context = context or {}
+        tempExpression = self.tokens
```
Schamper:
```suggestion
        temp_expression = self.tokens
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+        opKeys = set(self.operators.keys())
```
Schamper:
```suggestion
        op_keys = set(self.operators.keys())
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
 
-        return self.evaluate_part(buf, context)
+    def parseErr(self, condition: bool, error: str) -> None:
+        if condition:
+            raise ExpressionParserError(error)
 
-    def evaluate_part(self, buf: str, context: Dict[str, int]) -> int:
-        buf = buf.strip()
+    def evaluate(self, context: Dict[str, int] = None) -> int:
+        """Evaluates an expression using a Shunting-Yard implementation"""
```
Schamper:
```suggestion
        """Evaluates an expression using a Shunting-Yard implementation."""
```

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,284 @@
...
+            elif currentToken in context:
+                self.queue.append(context[currentToken])
+            elif currentToken in self.cstruct.consts:
+                self.queue.append(self.cstruct.consts[currentToken])
+            elif currentToken == "u":
+                self.stack.append(currentToken)
+            elif currentToken == "~":
+                self.stack.append(currentToken)
+            elif currentToken == "sizeof":
+                assert tempExpression[i + 1] == "("
```
Schamper: Can you change these assert's to exceptions?

Assert's get removed when running Python with the `-O` flag.

Schamper (changes requested):

sMezaOrellana:
Hi I see that the changes were just added. I was busy rounding up some uni work the past few weeks. Glad that this still found a place in the project.

Schamper:
> Hi I see that the changes were just added. I was busy rounding up some uni work the past few weeks. Glad that this still found a place in the project.

No worries, I understand! I'd very much like this to be a part of the project so I found some free time to make my suggested changes myself.

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,291 @@
...
+            if self.match(self.operator):
+                continue
+
+            # if token is a single digit, keep looping over expression and build the number
+            elif self.match(self.digit, consume=False, append=False):
+                token += self.get_token()
+                self.consume()
+
+                # support for binary and hexadecimal notation
+                if self.match(expected={"x", "X", "b", "B"}, consume=False, append=False):
```
cecinestpasunepipe: Seems to break on hex with B: "0x1B"

tests/test_expression.py
```diff
@@ -44,13 +44,19 @@
     ("0 | 1", 1),
     ("1 | 1", 1),
     ("1 | 2", 3),
-    # This type of expression is not supported by the parser and will fail.
-    # ('4 * 1 + 1', 5),
+    ("4 * 1 + 1", 5),
     ("-42", -42),
```
cecinestpasunepipe, line 52:
```suggestion
    ("-42", -42),
    ("0x1B", 0x1B),
```

cecinestpasunepipe (changes requested):

dissect/cstruct/expression.py
```diff
@@ -111,8 +112,8 @@ def tokenize(self) -> list[str]:
                 else:
                     pass
 
-                # number cannot end on x or b in the case of binary or hexadecimal notation
-                if token[-1] in ("x", "X", "b", "B"):
+                # number cannot end on x in the case of binary or hexadecimal notation
+                if token[-1] in ("x", "X"):
```
sMezaOrellana: This is to check that '0b' is not tokenized
```
if len(token) == 2 and  token[-1] in ("x", "X", "b", "B"):
```
This should fix it properly. I would commit it myself but I'm currently on holiday

Schamper: Awesome, thank you!

tests/test_expression.py
```diff
@@ -44,13 +45,21 @@
     ("0 | 1", 1),
     ("1 | 1", 1),
     ("1 | 2", 3),
```
cecinestpasunepipe, line 47:
```suggestion
    ("1 | 2", 3),
    ("1 | 2 | 4", 7),
```
Chaining multiple bitwise operators  ("1 | 2 | 4", 7) seems to fail...

cecinestpasunepipe (changes requested):

tests/test_expression.py
Schamper: Fixed, I also fixed the precedence of the `&` operator which seemed to be wrong.

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,298 @@
...
+                raise ExpressionParserError("Invalid expression: not enough operands")
+
+            left = self.queue.pop(-1)
+            res = self.operators[operator](left, right)
 
-            buf += self.expression[i]
+        self.queue.append(res)
+
+    def is_number(self, token: str) -> bool:
+        return token.isnumeric() or (len(token) > 2 and token[0] == "0" and token[1] in ("x", "X", "b", "B", "o", "O"))
```
cecinestpasunepipe, line 221: This can still give problems with notations like 09 although that is not very pretty notation anyway. However the error will be quite cryptic: `ValueError: invalid literal for int() with base 0: '0o9'`
Probably better to allow leading zeroes or detect the invalid notation correctly like Python itself:
`SyntaxError: leading zeros in decimal integer literals are not permitted; use an 0o prefix for octal integers`

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,298 @@
...
+            if self.match(self.operator):
+                continue
+
+            # If token is a single digit, keep looping over expression and build the number
+            elif self.match(self.digit, consume=False, append=False):
+                token += self.get_token()
+                self.consume()
+
+                # Support for binary and hexadecimal notation
+                if self.match(expected={"x", "X", "b", "B"}, consume=False, append=False):
```
cecinestpasunepipe: This breaks on octal notation, 0o777 etc.. Maybe put this dict in a const also, otherwise you have to update it in multiple places.

tests/test_expression.py
```diff
@@ -44,13 +45,24 @@
...
+    ("1 | 2 | 4", 7),
+    ("1 & 1 * 4", 0),
+    ("(1 & 1) * 4", 4),
+    ("4 * 1 + 1", 5),
     ("-42", -42),
     ("42 + (-42)", 0),
     ("A + 5", 13),
     ("21 - B", 8),
     ("A + B", 21),
+    ("~1", -2),
```
cecinestpasunepipe, line 57:
```suggestion
    ("0o77 + 0o66", 117),
    ("~1", -2),
```

Maybe add some cases for octal?

cecinestpasunepipe (changes requested):

dissect/cstruct/expression.py
Schamper: Leading `0` is generally interpreted as being octal notation, so the error is correct. GCC will raise a similar error:
```
x.c:4:11: error: invalid digit '9' in octal constant
        int a = 09;
```
However, putting e.g. `07` will work fine because that's a valid octal.

dissect/cstruct/expression.py
Schamper: This does not seem to be valid C?
```
x.c:4:11: error: invalid suffix 'o7' on integer constant
        int a = 0o7;
```

sMezaOrellana: > This breaks on octal notation, 0o777 etc.. Maybe put this dict in a const also, otherwise you have to update it in multiple places.

0o777 is not valid C. We parse an octal number in C and convert it to a Python-style octal notation.

```
                if len(token) > 1 and token[0] == "0" and token[1] not in ("x", "X", "b", "B"):
                    token = token[:1] + "o" + token[1:]
```
I hope this adds some clarity unless I misunderstood something.

cecinestpasunepipe: Ah ok, my bet, I thought this expression parser allowed Python-style expressions, but it makes sense to stick to C.

Schamper: Does that also address your other comments?

cecinestpasunepipe: Yes, thanks. I am currently test-driving this branch while working on things. So far it seems to work well.

Schamper: Alright, I've marked your other comments as resolved. Let me know once you're satisfied!

dissect/cstruct/expression.py
```diff
@@ -1,84 +1,298 @@
...
+                    self.match(expected={"l", "L"}, append=False)
+
+                elif self.match(expected={"l", "L"}, append=False):
+                    self.match(expected={"l", "L"}, append=False)
+                    self.match(expected={"u", "U"}, append=False)
+                else:
+                    pass
+
+                # Number cannot end on x or b in the case of binary or hexadecimal notation
+                if len(token) == 2 and token[-1] in ("x", "X", "b", "B"):
```
cecinestpasunepipe: Maybe I did not understand the response to my previous question about this, but isn't it an idea to use a constant here?

i.e.

HEXBIN_SUFFIX = ("x", "X", "b", "B")

?

cecinestpasunepipe (changes requested):

dissect/cstruct/expression.py
Schamper: Oh I forgot about that, apologies.

cecinestpasunepipe (approved):

Schamper (approved):

Schamper:
@sMezaOrellana again, huge thanks for this PR!

Issue #30, opened 2023-04-12
Anonymous enums


d3dave:
cstruct fails to parse anonymous enums because the regex requires enums to have a name. We should be able to just treat all enum members as top-level defines.

Schamper:
Good suggestion. The problem is currently twofold: 
- The regex + parsing for enums needs to allow anonymous enums
- Enum values need to be added as constants to the global scope. Currently only the enum name is added and values need to be accessed as `Name.Value`.

Schamper:
This has now been implemented with https://github.com/fox-it/dissect.cstruct/pull/39.

PR #39, opened 2023-06-17
Anonymous enums

sMezaOrellana:
Add support for anonymous enums (https://github.com/fox-it/dissect.cstruct/issues/30). Anonymous enums are treated as top-level defines.

Changed the tokenizer to make the name inside an enum optional.

`r"(?P<enumtype>enum|flag)\s+(?P<name>[^\s:{]+)?\s*(:\s"`

If the enum does not contain a name and the enumtype equals enum we add the key and values of the enum to the consts dictionary.
One thing that I think is now possible is the following
```
flag {
    a,
    b,
    c
}
```
Not sure what the behavior of an anonymous flag should be. If not allowed the only change should be to throw an error if `enumtype == flag and not d["name"]`

Added some tests as well

### DIFF
--- a/dissect/cstruct/expression.py
+++ b/dissect/cstruct/expression.py
@@ -79,6 +79,6 @@ class Expression:
             return int(buf, 16)
 
         if buf in self.cstruct.consts:
-            return self.cstruct.consts[buf]
+            return int(self.cstruct.consts[buf])
 
         return int(buf)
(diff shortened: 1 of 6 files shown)


Schamper:
I don't necessarily see any issues with allowing anonymous `flag`'s, so perhaps just allow it? If you could also add a unit tests for that, that would be great :)

Schamper:
I've made a few changes, one of which is expanding to the flag type, as well as improving the consistency by ensuring that all anonymous members are actually instances of the enum, instead of just their numeric value.

pyrco (approved):

PR #42, opened 2023-07-28
Allow enums to reference their own members

### LINKED ISSUE #41: Cannot use previous enum constants within expressions for following values
If we have an enum like this:

```c
enum test {
  A,
  B,
  C = B,
  D
};
```

we get an error while parsing because `B` is not within the scope yet (all of the constants are added to the global scope only after the entire enum has finished parsing).

Schamper:
Closes https://github.com/fox-it/dissect.cstruct/issues/41

### DIFF
--- a/tests/test_enum.py
+++ b/tests/test_enum.py
@@ -260,3 +260,19 @@ def test_enum_anonymous_struct(compiled):
 
     t = test(b"\xff\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x0A\x00\x00\x00")
     assert t.arr == [255, 0, 0, 10]
+
+
+def test_enum_reference_own_member(compiled):
+    cdef = """
+    enum test {
+        A,
+        B = A + 3,
+        C
+    };
+    """
+    cs = cstruct.cstruct()
+    cs.load(cdef, compiled=compiled)
+
+    assert cs.test.A == 0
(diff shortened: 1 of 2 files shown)


tests/test_enum.py
```diff
@@ -260,3 +260,19 @@ def test_enum_anonymous_struct(compiled):
...
     t = test(b"\xff\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x0A\x00\x00\x00")
     assert t.arr == [255, 0, 0, 10]
+
+
+def test_enum_reference_own_member(compiled):
+    cdef = """
+    enum test {
+        A,
+        B = A + 3,
+        C
```
Miauwkeru, line 270: Just a small question, if you create a struct like this:
```python
enum test {
    A = 1,      // 1
    B = A + 3,  // 3
    C = B + A,  // 3
    D = B - C,  // 0
    E = D + D,  // 0
    G
};
```
`G` would have a value of `-1` however, that value is already available in `test`, and is negative while the type of the enum is uint32... is this intentional behavior?

Schamper: UB :smile:

In all seriousness, this is already an "issue" because you can just do `A = -1`. I don't think this is something we need to address at all, let alone in this PR.

Miauwkeru (approved):

PR #43, opened 2023-08-02
Preserve newlines when removing comments

### LINKED ISSUE #31: Stripping comments does not preserve line numbers
If there are any parse errors after multi-line comments, the reported line numbers for the errors are incorrect. To fix this, we can preserve any newlines within comments.

Schamper:
So that the line number remains accurate.

Also, actually fix the line number to be accurate :) it was off by one.

Fixes #31

### DIFF
--- a/dissect/cstruct/parser.py
+++ b/dissect/cstruct/parser.py
@@ -294,16 +294,16 @@ class TokenParser(Parser):
     @staticmethod
     def _remove_comments(string: str) -> str:
         # https://stackoverflow.com/a/18381470
-        pattern = r"(\".*?\"|\'.*?\')|(/\*.*?\*/|//[^\r\n]*$)"
+        pattern = r"([\"\'].*?[\"\'])|(/\*.*?\*/|//[^\r\n]*$)"
         # first group captures quoted strings (double or single)
         # second group captures comments (//single-line or /* multi-line */)
         regex = re.compile(pattern, re.MULTILINE | re.DOTALL)
 
-        def _replacer(match):
+        def _replacer(match: re.Match) -> str:
             # if the 2nd group (capturing comments) is not None,
             # it means we have captured a non-quoted (real) comment string.
-            if match.group(2) is not None:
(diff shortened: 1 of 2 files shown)


dissect/cstruct/parser.py
```diff
@@ -294,16 +294,16 @@ def _names(self, tokens: TokenConsumer) -> List[str]:
     @staticmethod
     def _remove_comments(string: str) -> str:
         # https://stackoverflow.com/a/18381470
-        pattern = r"(\".*?\"|\'.*?\')|(/\*.*?\*/|//[^\r\n]*$)"
+        pattern = r"([\"\'].*?[\"\'])|(/\*.*?\*/|//[^\r\n]*$)"
```
Poeloe: It now also triggers on this `"<anything other than ' or ">'` situation or vice versa, is that what you wanted to include here?

So to elaborate, if someone uses the comment: `"This is a 'tricky' comment"` the first group will be `"This is a '` since `.*` is not greedy.

Schamper: It's indeed different, reverted the pattern.

tests/test_parser.py
```diff
@@ -0,0 +1,26 @@
+from unittest.mock import Mock
+
+from dissect.cstruct.parser import TokenParser
+
+
+def test_preserve_comment_newlines():
```
Poeloe, line 6: Maybe include the other comment options (since they are included in the pattern) also in the test and add the situations `"foo 'bar'"` and `'foo "bar"'`.

Poeloe (changes requested):

tests/test_parser.py
Schamper: That's not what you think it is. That regex is to leave "comments" that exist within a string alone. So e.g. `// comment` gets deleted but `"oh wowie // is such a cool operator"` doesn't.

Poeloe: Aah okay, never mind then!

Poeloe (approved):
LGTM!

PR #35, opened 2023-05-31
Rewrite type system and compiler (v4)

Schamper:
- Rewrite entire type system to be class based
  - All types are now classes, and parsed values are instances of that class
- Rewrite compiler to be compatible with this type system

TODO:

- [x] Add unit test for custom type
- [ ] Add example for custom type
- [ ] Determine repr() for structure class (not instance)
- [ ] Work out pickling?
  - [ ] Maybe related, but work out "module path" for new type classes
- [ ] Update README

(DIS-75, DIS-77, DIS-103, DIS-104, DIS-965, DIS-1087, DIS-1124, DIS-1970)

PR #44, opened 2023-08-17
Union processing fix proposed

andrik75:
Currently a union instance saves its data by taking the longest field in the union. Bit it doesn't guarantee it's proper data after some field has been modified there. The fix fixes the issue.

### DIFF
--- a/dissect/cstruct/types/instance.py
+++ b/dissect/cstruct/types/instance.py
@@ -7,13 +7,24 @@ from dissect.cstruct.types import BaseType
 class Instance:
     """Holds parsed structure data."""
 
-    __slots__ = ("_type", "_values", "_sizes")
+    __slots__ = ("_type", "_values", "_sizes", "owner")
 
     def __init__(self, type_: BaseType, values: Dict[str, Any], sizes: Dict[str, int] = None):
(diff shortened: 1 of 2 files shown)


andrik75:
Currently a union instance saves its data by taking the longest field in the union. But it doesn't guarantee it's proper data after some field has been modified there. The fix fixes the issue.

Schamper:
Wanted to give a short status update here. I've been trying out my idea of having "proxy" objects only for unions to hopefully not impact performance too much for other structures. I've got it mostly working, just need to figure out a solution for anonymous unions. If that doesn't work out, I'll probably end up making some suggestions to your PR that ends up being somewhere in the middle.

Schamper:
Another small update. I've decided to only work on adding this feature to the upcoming cstruct "v4" (#35). The way how I wanted to incorporate this is a lot easier with the new type model introduced there.

You can view my solution in the linked PR (#49). It would be interesting to learn if all of your use-cases are covered by that branch.

Schamper:
#49 has since been merged into the v4 branch. It won't be long until we merge that into main. I'll close this PR but would like to thank you for the idea and suggestion!

PR #49, opened 2023-10-29
Add support for real-time updating of union  (v4)

### LINKED ISSUE #48: Issue when there is union inside structure
suppose there is a structure  and inside that there is union so when i am assigning value to the first variable in union then it show in binary file but if i assign value to second variable in union then in binary file the value is not showing, so can you check why this is happening

Schamper:
This mainly adds support for the real-time updating of union members (also when nested deep in structures within the union) as described in https://github.com/fox-it/dissect.cstruct/pull/44 and https://github.com/fox-it/dissect.cstruct/issues/48.

In the process, this also changes some things with how anonymous fields are managed, as well as renaming some structure metadata fields.

Closes #48 and supersedes #44. Obviously depends on #35.

### DIFF
--- a/dissect/cstruct/types/structure.py
+++ b/dissect/cstruct/types/structure.py
@@ -535,7 +535,7 @@ class Union(Structure, metaclass=UnionMetaType):
         object.__setattr__(self, "_sizes", sizes)
 
     def _proxify(self) -> None:
-        def _proxy_structure(value: Structure) -> UnionProxy:
+        def _proxy_structure(value: Structure) -> None:
             for field in value.__class__.__fields__:
                 if issubclass(field.type, Structure):
                     nested_value = getattr(value, field.name)
--- a/dissect/cstruct/types/structure.py
+++ b/dissect/cstruct/types/structure.py
@@ -33,12 +33,16 @@ class StructureMetaType(MetaType):
 
     # TODO: resolve field types in _update_fields, remove resolves elsewhere?
 
-    fields: list[Field]
+    fields: dict[str, Field]
+    """Mapping of field names to :class:`Field` objects, including "folded" fields from anonymous structures."""
     lookup: dict[str, Field]
+    """Mapping of "raw" field names to :class:`Field` objects. E.g. holds the anonymous struct and not its fields."""
+    __fields__: list[Field]
+    """List of :class:`Field` objects for this structure. This is the structures' Single Source Of Truth."""
 
-    __anonymous__: bool
+    # Internal
     __align__: bool
-    __lookup__: dict[str, Field]
+    __anonymous__: bool
     __updating__ = False
     __compiled__ = False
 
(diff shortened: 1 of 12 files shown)


dissect/cstruct/types/structure.py
```diff
@@ -456,9 +503,90 @@ def _write(cls, stream: BinaryIO, data: Union) -> int:
...
+    def _proxify(self) -> UnionProxy:
+        def _proxy_structure(value: Structure) -> UnionProxy:
+            for field in value.__class__.__fields__:
+                if issubclass(field.type, Structure):
+                    nested_value = getattr(value, field.name)
+                    proxy = UnionProxy(self, field.name, nested_value)
+                    object.__setattr__(value, field.name, proxy)
+                    _proxy_structure(nested_value)
+
+        _proxy_structure(self)
```
pyrco, line 546: There are no `return`s here, but I guess they're not needed, so the return types should become `None`.

dissect/cstruct/types/structure.py
```diff
@@ -65,29 +77,39 @@ def _update_fields(
...
-        classdict["__lookup__"] = raw_lookup
-        classdict["__init__"] = _patch_args_and_attributes(_make__init__(num_fields), field_names)
+        init_names = raw_lookup.keys()
+        classdict["fields"] = lookup
+        classdict["lookup"] = raw_lookup
+        classdict["__fields__"] = fields
         classdict["__bool__"] = _patch_attributes(_make__bool__(num_fields), field_names, 1)
 
-        if issubclass(cls, UnionMetaType):
+        if issubclass(cls, UnionMetaType) or isinstance(cls, UnionMetaType):
```
pyrco, line 105:
```suggestion
        if issubclass(cls, UnionMetaType) or isinstance(cls, UnionMetaType):
```

`issubclass` will raise an exception if `cls` is an instance, so this needs to be guarded.

pyrco (changes requested):

dissect/cstruct/types/structure.py
Schamper: `cls` is always a class. The `isinstance()` here checks if it's an instance of a `UnionMetaType`, which is a class.

dissect/cstruct/types/structure.py
```diff
@@ -456,9 +503,90 @@ def _write(cls, stream: BinaryIO, data: Union) -> int:
...
+        self._update()
+
+    def _update(self) -> None:
+        result, sizes = self.__class__._read_fields(io.BytesIO(self._buf))
+        self.__dict__.update(result)
+        object.__setattr__(self, "_values", result)
+        object.__setattr__(self, "_sizes", sizes)
+
+    def _proxify(self) -> None:
+        def _proxy_structure(value: Structure) -> UnionProxy:
```
pyrco:
```suggestion
        def _proxy_structure(value: Structure) -> None:
```

dissect/cstruct/types/structure.py
pyrco: Ah, because it is part of the metaclass apparently, check.

pyrco (changes requested):

dissect/cstruct/types/structure.py
Schamper: I had to double check as well, but this code path has unit tests for both conditions so I can at least assure it works :smile:

pyrco (approved):

PR #50, opened 2023-10-30
Fix dumpstruct for anonymous structures

Schamper:


tests/test_util.py
```diff
@@ -23,32 +23,41 @@ def test_hexdump(capsys):
 
 
 def test_dumpstruct(capsys, compiled):
```
Miauwkeru:
```suggestion
@pytest.mark.parametrize(
    "cdef",
    [
        """
    struct test {
        uint32 testval;
    };
    """,
        """
    struct test {
        struct {
            uint32 testval;
        };
    };
    """,
    ],
)
def test_dumpstruct(capsys, compiled: bool, cdef: str) -> None:
```
wouldn't this be a bit neater?

Miauwkeru (changes requested):

tests/test_util.py
Schamper: I don't find this very readable, plus we may want to test more complex dumpstructs later that don't follow this simple uint32 test value.

Miauwkeru: Fair, tho when running the test. if one of them fails you'd need to check which one if them fails first. If you separate them in different tests, you at least have an idea where to look at/which functionality broke.

as for the future dumpstruct tests, wouldn't you just add separate tests for those anyways? And if the check logic is the same, maybe move the check logic to its own function or something.

Schamper: Put this in a separate test

Miauwkeru (approved):
LGTM

PR #53, opened 2023-11-24
Return self on load for chaining

Schamper:
Makes the API a bit nicer for loading definitions.

```python
c_foo = cstruct().load(foo_def)
```


yunzheng (approved):
LGTM

PR #58, opened 2023-12-09
data-backed unary operator selection (v4)

xloem:
Does this look right to you?

I noticed the unary operators were hardcoded, so pulled them out so more could be added if needed.  (A useful unary operator is for example to extract the value of single-element arrays.)

A more powerful solution could be to use an operator dataclass, although this would increase the unmerged changes further.

### DIFF
--- a/dissect/cstruct/expression.py
+++ b/dissect/cstruct/expression.py
@@ -154,7 +154,7 @@ class ExpressionTokenizer:
 class Expression:
     """Expression parser for calculations in definitions."""
 
-    operators = {
+    binary_operators = {
         "|": lambda a, b: a | b,
         "^": lambda a, b: a ^ b,
         "&": lambda a, b: a & b,
@@ -165,6 +165,9 @@ class Expression:
         "*": lambda a, b: a * b,
         "/": lambda a, b: a // b,
         "%": lambda a, b: a % b,
+    }
+
+    unary_operators = {
         "u": lambda a: -a,
         "~": lambda a: ~a,
     }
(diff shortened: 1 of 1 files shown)


xloem:
Note: the operator code could possibly be simplified away entirely into a call to python’s expression `eval` function.

Poeloe:
Hi @xloem, thanks for your contribution!
I will have a look at the code as soon as possible and will get back to you. Stay tuned 😉 

PS: don't forget to agree to the CLA

xloem:
Been a long time since I’ve seen a CLA. It saddens me that the license was changed from AGPL to Apache without comment or explanation in 2022.
@DissectBot agree [company="{default}"]

Poeloe (approved):
Code looks good, thanks for you addition. I will merge it soon.

**About the `eval` suggestion:** 
We decided to use a "safer" expression evaluation, since `eval` has its own set of security pitfalls. Something like `ast.literal_eval` (https://docs.python.org/3/library/ast.html#ast.literal_eval) would be preferred, but that also comes with its own restrictions and pitfalls.

If you happen to have any further suggestions on this topic, we're more than happy to hear them!

xloem:
Thanks for the merge !

> **About the `eval` suggestion:** We decided to use a "safer" expression evaluation, since `eval` has its own set of security pitfalls. Something like `ast.literal_eval` (https://docs.python.org/3/library/ast.html#ast.literal_eval) would be preferred, but that also comes with its own restrictions and pitfalls.
> 
> If you happen to have any further suggestions on this topic, we're more than happy to hear them!

If a symbolic algebra system were used for expressions here it might pave the way for automatically backsolving to derive some field values like array lengths when writing.

PR #59, opened 2023-12-09
Bugfix/#56 missing f string

### LINKED ISSUE #56: Field __repr__ missing "f" qualifier on the f-string
https://github.com/fox-it/dissect.cstruct/blob/ffbf4f83169d5c6db2448d42c0e2fa5cec5c874a/dissect/cstruct/types/structure.py#L25

Before:
`<Field res0 uint16 : {self.bits}>`

After making it an f-string:
`<Field res0 uint16 : 6>`

I think the fix is as easy as that. Make line 25 use an f-string. Though I'd appreciate if someone could confirm/deny this is the correct solution for this codebase.

burneykb:
Fixes #56 

I was messing around and I saw that the `Field` class was missing the "f" in it's `__repr__`. This small PR adds the missing "f" and also adds linting rules to catch this in the future.

According to the linter there were no additional instances of this mistake in the repo. However there was one false positive in the `Compiler` class. In this case the missing "f" is intentional and as such I've ignored the linting rule on the single line. Of course this caused the line to be longer than 120 characters.. so I've also ignored the line length rule as well. I'm not super familiar with linters so please let me know if there is a better way to do this. :)

### DIFF
--- a/dissect/cstruct/compiler.py
+++ b/dissect/cstruct/compiler.py
@@ -280,7 +280,7 @@ class {name}(Structure):
                 cur_type = pack_char
 
             if isinstance(read_type, BytesInteger):
-                getter = "BytesInteger.parse(buf[{slice}], {size}, {count}, {signed}, self.cstruct.endian){data_slice}"  # noqa: FS003, E501
+                getter = "BytesInteger.parse(buf[{slice}], {size}, {count}, {signed}, self.cstruct.endian){data_slice}"
                 getter = getter.format(
                     slice=read_slice,
                     size=read_type.size,
--- a/dissect/cstruct/compiler.py
+++ b/dissect/cstruct/compiler.py
@@ -281,6 +281,7 @@ class {name}(Structure):
 
             if isinstance(read_type, BytesInteger):
                 getter = "BytesInteger.parse(buf[{slice}], {size}, {count}, {signed}, self.cstruct.endian){data_slice}"
+
                 getter = getter.format(
                     slice=read_slice,
                     size=read_type.size,
--- a/dissect/cstruct/compiler.py
+++ b/dissect/cstruct/compiler.py
@@ -280,8 +280,7 @@ class {name}(Structure):
                 cur_type = pack_char
 
             if isinstance(read_type, BytesInteger):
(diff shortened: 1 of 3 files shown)


burneykb:
@DissectBot ,

@DissectBot agree [company="default"]

burneykb:
@DissectBot agree [company="default"]

tox.ini
```diff
@@ -43,10 +43,11 @@ deps =
     flake8
     flake8-black
     flake8-isort
+    flake8-use-fstring
```
Schamper: @pyrco thoughts on adding this?

pyrco: I think we shouldn't add this. We keep the number of flake8 extensions limited on purpose, as they often have unwanted side-effects. This one included as it needs `--enable-extensions=FS003` to catch this type of bug, which may trigger on legitimate `{}` in strings that er not f-strings or are templates.

dissect/cstruct/compiler.py
```diff
@@ -280,8 +280,7 @@ def gen_read_block(self, size: int, block: List[str]) -> str:
                 cur_type = pack_char
 
             if isinstance(read_type, BytesInteger):
-                getter = "BytesInteger.parse(buf[{slice}], {size}, {count}, {signed}, self.cstruct.endian){data_slice}"
-
+                getter = "BytesInteger.parse(buf[{slice}], {size}, {count}, {signed}, self.cstruct.endian){data_slice}"  # noqa: FS003, E501
```
Schamper:
```suggestion
                getter = "BytesInteger.parse(buf[{slice}], {size}, {count}, {signed}, self.cstruct.endian){data_slice}"
```

tox.ini
Schamper:
```suggestion
```

tox.ini
```diff
@@ -43,10 +43,11 @@ deps =
...
     isort==5.11.4
     vermin
 commands =
-    flake8 dissect tests
+    flake8 dissect tests --enable-extensions=FS003
```
Schamper:
```suggestion
    flake8 dissect tests
```

Schamper (changes requested):

tox.ini
burneykb: Alrighty, I'll get rid of these changes then. 👍

Schamper (approved):

PR #60, opened 2023-12-09
Data-Backed Array Types (v4)

xloem:
I had made a custom type based on the main branch to include values conditionally based on array length. This failed in v4 because the result is passed to the array constructor which is selected in a hardcoded manner. This change parameterizes the array type as a class attribute.

Here’s my new prototype conditional type using this patch:
```
class ConditionalType(BaseType):
    type : BaseType
    dflt : Any
    class ArrayType(BaseType, metaclass=ArrayMetaType):
        @classmethod
        def _read(cls, stream, context):
            count = cls.num_entries.evaluate(context)
            assert count == TRUE or count == FALSE
            if count == TRUE:
                return cls.type.type._read(stream, context)
            else:
                return cls.type.dflt
        @classmethod
        def _write(cls, stream, data):
            if data is not None and data != cls.dflt:
                return cls.type.type._write(stream, data) 
            else:
                return 0
```


dissect/cstruct/types/base.py
```diff
@@ -217,3 +217,4 @@ def _write(cls, stream: BinaryIO, data: list[Any]) -> int:
             raise ArraySizeError(f"Expected static array size {cls.num_entries}, got {actual_size} instead.")
 
         return cls.type._write_array(stream, data)
+BaseType.ArrayType = Array
```
Schamper, line 220: Could you add a `ArrayType = None` to the `BaseType` class with a short comment describing what it's for? Something along the lines of:

```
# Define the array type for this type, this defaults to the Array class below and is set after that's defined
ArrayType: ArrayMetaType = None
```

Schamper (changes requested):
Could you also fix the linting errors and add a small unit test with a custom `ArrayType`?

xloem:
Thanks for your reviews. I hope to get to this after the holidays.

Schamper:
I've rebased locally and fixed the issues, but unfortunately I'm not allowed to force-push to your branch. @xloem can you please indicate if you will work on this soon? Otherwise I'll manually put these changes into the v4 branch.

xloem:
I’m so sorry, I’m no longer on this for now. I may return in future months. Thank you so much for engaging it. Feel free to close or to manually merge.

Schamper:
No problem at all, I value the contributions! Looking forward to what else you may have in store in the future.

PR #61, opened 2023-12-14
Field access to outer structure scope (v4)

xloem:
This is an attempt to provide access to outer fields within structure expressions, such that the following is valid:
```
struct {
  char str[len];
} len_str;
struct {
  int32 len;
  len_str str;
} obj;
```
Locally, I have some lengths in headers that are then used inside array objects.


Schamper:
Could you add some unit tests, perhaps also with some more complex cases such as conflicting field names?

The v4 refactor branch also had some updates (a PR was merged into it), could you rebase?

Schamper:
After some deliberation I've decided not to include this functionality. I feel like it's too error prone and susceptible to weird edge cases. A more fleshed out version (e.g. through a reserved keyword, akin to `super()`) may be considered in the future, but for now I'll close this.

Thank you for your efforts though!

PR #69, opened 2024-02-15
Add LEB128 variable-lengh integer support

joost-j:
During the development of a `dissect.target` plugin for parsing Windows Notepad tabs (which will be submitted soon as well), I stumbled upon a variable-length integer encoding scheme that was not yet present in dissect. All is well explained on Wikipedia: https://en.wikipedia.org/wiki/LEB128. This implementation is valid for both signed- and unsigned LEB128 integers, the Wikipedia pseudo-code and some reference implementations on the internet were used to check the code.

Targeting the `refactor-v4`  for now, so that it can hopefully be shipped with the next major release!

### DIFF
--- a/dissect/cstruct/cstruct.py
+++ b/dissect/cstruct/cstruct.py
@@ -29,6 +29,7 @@ from dissect.cstruct.types import (
     Wchar,
     WcharArray,
 )
+from dissect.cstruct.types.leb128 import LEB128
 
 
 class cstruct:
@@ -70,6 +71,9 @@ class cstruct:
             "uint48": self._make_int_type("int48", 6, False, alignment=8),
             "int128": self._make_int_type("int128", 16, True, alignment=16),
             "uint128": self._make_int_type("uint128", 16, False, alignment=16),
+
+            "uleb128": self._make_type('uleb128', (LEB128,), None, alignment=4, attrs={"signed": False}),
+            "ileb128": self._make_type('ileb128', (LEB128,), None, alignment=4, attrs={"signed": True}),
 
             "void": self._make_type("void", (Void,), 0),
 
--- a/dissect/cstruct/cstruct.py
+++ b/dissect/cstruct/cstruct.py
@@ -11,6 +11,7 @@ from dissect.cstruct.exceptions import ResolveError
 from dissect.cstruct.expression import Expression
 from dissect.cstruct.parser import CStyleParser, TokenParser
 from dissect.cstruct.types import (
+    LEB128,
     Array,
     ArrayMetaType,
     BaseType,
@@ -29,7 +30,6 @@ from dissect.cstruct.types import (
     Wchar,
     WcharArray,
 )
-from dissect.cstruct.types.leb128 import LEB128
 
 
 class cstruct:
--- a/dissect/cstruct/types/leb128.py
+++ b/dissect/cstruct/types/leb128.py
@@ -20,7 +20,7 @@ class LEB128(int, BaseType):
         while True:
             b = stream.read(1)
             if b == b"":
-                raise EOFError("EOF reached, while final LEB128 byte was not yet read.")
+                raise EOFError("EOF reached, while final LEB128 byte was not yet read")
 
             b = ord(b)
             result |= (b & 0x7F) << shift
@@ -50,7 +50,7 @@ class LEB128(int, BaseType):
     def _write(cls, stream: BinaryIO, data: int) -> LEB128:
         # only write negative numbers when in signed mode
         if data < 0 and not cls.signed:
-            raise ValueError("Attempt to encode a negative integer using unsigned LEB128 encoding.")
+            raise ValueError("Attempt to encode a negative integer using unsigned LEB128 encoding")
 
         result = bytearray()
         while True:
--- a/dissect/cstruct/types/leb128.py
+++ b/dissect/cstruct/types/leb128.py
@@ -50,7 +50,7 @@ class LEB128(int, BaseType):
     def _write(cls, stream: BinaryIO, data: int) -> LEB128:
         # only write negative numbers when in signed mode
         if data < 0 and not cls.signed:
-            return 0
+            raise ValueError("Attempt to encode a negative integer using unsigned LEB128 encoding.")
 
         result = bytearray()
         while True:
--- a/dissect/cstruct/types/leb128.py
+++ b/dissect/cstruct/types/leb128.py
@@ -7,7 +7,9 @@ from dissect.cstruct.types.base import BaseType
 
 class LEB128(int, BaseType):
     """Variable-length code compression to store an arbitrarily large integer in a small number of bytes.
-    See https://en.wikipedia.org/wiki/LEB128 for more information and an explanation of the algorithm."""
+
+    See https://en.wikipedia.org/wiki/LEB128 for more information and an explanation of the algorithm.
+    """
 
     signed: bool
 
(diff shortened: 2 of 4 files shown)


dissect/cstruct/cstruct.py
```diff
@@ -29,6 +29,7 @@
     Wchar,
     WcharArray,
```
Schamper: Can you add the LEB128 to `dissect/cstruct/types/__init__.py` and then add it to this import list?

dissect/cstruct/types/leb128.py
```diff
@@ -0,0 +1,27 @@
+from __future__ import annotations
+
+from typing import Any, BinaryIO
+
+from dissect.cstruct.types.base import BaseType
+
+
+class LEB128(int, BaseType):
+    """Variable-length code compression to store an arbitrarily large integer in a small number of bytes.
+    See https://en.wikipedia.org/wiki/LEB128 for more information and an explanation of the algorithm."""
```
Schamper:
```suggestion
    """Variable-length code compression to store an arbitrarily large integer in a small number of bytes.
    
    See https://en.wikipedia.org/wiki/LEB128 for more information and an explanation of the algorithm.
    """
```

dissect/cstruct/types/leb128.py
```diff
@@ -0,0 +1,27 @@
...
+
+    signed: bool
+
+    @classmethod
+    def _read(cls, stream: BinaryIO, context: dict[str, Any] = None) -> LEB128:
```
Schamper, line 17: Can you also add write support? :smile:

tests/test_types_leb128.py
```diff
@@ -0,0 +1,96 @@
...
+def test_leb128_signed_read(cs: cstruct):
+    assert cs.ileb128(b"\x02") == 2
+    assert cs.ileb128(b"\x8b%") == 4747
+    assert cs.ileb128(b"\xc9\x8f\xb0\x06") == 13371337
+    assert cs.ileb128(b"~") == -2
+    assert cs.ileb128(b"\xf5Z") == -4747
+    assert cs.ileb128(b"\xde\xd6\xcf|") == -7083170
+
+
+def test_leb128_struct_unsigned(cs: cstruct, compiled: bool):
```
Schamper:
```suggestion
def test_leb128_struct_unsigned(cs: cstruct):
```
We won't be testing if the structure gets compiled, because it won't be supported because of the unsupported type. So you can remove all references of `compiled`.

Schamper (changes requested):

dissect/cstruct/types/leb128.py
```diff
@@ -0,0 +1,27 @@
...
+        while True:
+            b = ord(stream.read(1))
+            result |= (b & 0x7F) << shift
+            shift += 7
+            if (b & 0x80) == 0:
+                break
+        if cls.signed:
+            if b & 0x40 != 0:
+                result |= ~0 << shift
+        return result
```
Schamper:
```suggestion

        if cls.signed:
            if b & 0x40 != 0:
                result |= ~0 << shift

        return result
```

dissect/cstruct/types/leb128.py
```diff
@@ -0,0 +1,27 @@
...
+    See https://en.wikipedia.org/wiki/LEB128 for more information and an explanation of the algorithm."""
+
+    signed: bool
+
+    @classmethod
+    def _read(cls, stream: BinaryIO, context: dict[str, Any] = None) -> LEB128:
+        result = 0
+        shift = 0
+        while True:
+            b = ord(stream.read(1))
```
Schamper: What if we reach EOF? See the Int class for a similar way of dealing with that.

joost-j: Added better EOF handling + a test to check for this.

dissect/cstruct/types/leb128.py
joost-j: Done

dissect/cstruct/types/leb128.py
```diff
@@ -50,7 +50,7 @@ def _read_0(cls, stream: BinaryIO, context: dict[str, Any] = None) -> LEB128:
     def _write(cls, stream: BinaryIO, data: int) -> LEB128:
         # only write negative numbers when in signed mode
         if data < 0 and not cls.signed:
-            return 0
+            raise ValueError("Attempt to encode a negative integer using unsigned LEB128 encoding.")
```
Schamper:
```suggestion
            raise ValueError("Attempt to encode a negative integer using unsigned LEB128 encoding")
```

tests/test_types_leb128.py
```diff
@@ -154,3 +154,10 @@ def test_leb128_signed_write(cs: cstruct):
     assert cs.ileb128(-2).dumps() == b"~"
     assert cs.ileb128(-4747).dumps() == b"\xf5Z"
     assert cs.ileb128(-7083170).dumps() == b"\xde\xd6\xcf|"
+
+
+def test_leb128_write_negatives(cs: cstruct):
+    with pytest.raises(ValueError) as ex:
+        cs.uleb128(-2).dumps()
+    assert str(ex.value) == "Attempt to encode a negative integer using unsigned LEB128 encoding."
```
Schamper: You can use `pytest.raises(ValueError, match=...)` to match against the error text.

https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=raises#pytest.raises

tests/test_types_leb128.py
```diff
@@ -154,3 +154,10 @@ def test_leb128_signed_write(cs: cstruct):
...
+    assert cs.ileb128(-2).dumps() == b"~"
```
Schamper: Would you mind changing all byte strings to their raw "hex" escaped bytes? e.g. `b"\x7e"`.

Schamper (changes requested):

tests/test_types_leb128.py
joost-j: Done

tests/test_types_leb128.py
joost-j: Nice one!

dissect/cstruct/cstruct.py
```diff
@@ -29,6 +29,7 @@
...
 )
+from dissect.cstruct.types.leb128 import LEB128
```
Schamper: Can you still add the `LEB128` import into the big list just above this?

dissect/cstruct/types/leb128.py
```diff
@@ -0,0 +1,71 @@
...
+        while True:
+            if (value := cls._read(stream, context)) == 0:
+                break
+
+            result.append(value)
+
+        return result
+
+    @classmethod
+    def _write(cls, stream: BinaryIO, data: int) -> LEB128:
```
Schamper:
```suggestion
    def _write(cls, stream: BinaryIO, data: int) -> int:
```

dissect/cstruct/types/leb128.py
```diff
@@ -0,0 +1,71 @@
...
+            # the encoding process.
+            if (cls.signed and (data == 0 and byte & 0x40 == 0) or (data == -1 and byte & 0x40 != 0)) or (
+                not cls.signed and data == 0
+            ):
+                result.append(byte)
+                stream.write(result)
+                return len(result)
+
+            # Set high-order bit of byte
+            result.append(0x80 | byte)
```
Schamper, line 70: Missing a return of the amount of bytes written.

Schamper (changes requested):

dissect/cstruct/cstruct.py
joost-j: Ah, imported from the wrong file. Will fix it

dissect/cstruct/types/leb128.py
joost-j: It's returning `len(result)` in the `if` clause. in line 68. It just continues after line 71.

joost-j: I'll add a test for it

Schamper (approved):

PR #70, opened 2024-02-21
Add support for array and pointer typedefs

Schamper:
While the current syntax parser is quite restricting and due a rewrite, this will add support for a long time outstanding annoyance where the following syntax would not behave as expected:

```c
typedef char uuid_t[16];
```

Previously, this would simply add a type alias for `"char"` as the literal string `"uuid_t[16]"`. With this PR, this finally behaves as expected and allows you to typedef UUID character arrays!

### DIFF
--- a/dissect/cstruct/parser.py
+++ b/dissect/cstruct/parser.py
@@ -169,7 +169,7 @@ class TokenParser(Parser):
         for name in names:
             type_, name, bits = self._parse_field_type(type_, name)
             if bits is not None:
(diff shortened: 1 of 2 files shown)


yunzheng (approved):
Not sure if we also want to support multiple typedefs on a single line, for example:

```cpp
typedef char uuid_t[16], UUID[16];
```

If not, then it LGTM!

Edit: nevermind about the multiple typedefs on single line, i see that it's not part of this change :)

PR #71, opened 2024-02-21
Add support for pointer of pointer

Schamper:
Noticed this wasn't supported yet, small change to support it.


Schamper:
Noticed this doesn't actually work with the old compiler, and it's a pain to support it in there, so I'll close this and only make it available in v4.

https://github.com/fox-it/dissect.cstruct/pull/35/commits/0ef089a10e96404e0bb584aa54e8be9c28968b67

PR #72, opened 2024-02-27
Create a tool for stubs in cstruct v4

Miauwkeru:
The only thing that does not work as intended yet is the `dereference` of pointer. As in, it doesn't return the desired type information but just `T`


dissect/cstruct/tools/stubify.py
Schamper, line 1: Can you also add an `__init__.py` to this directory.

dissect/cstruct/tools/stubify.py
```diff
@@ -0,0 +1,105 @@
+# Searches and creates a stub of a cstruct definitions
```
Schamper: I'd put this as the description (maybe with a few more words) of the argument parser rather than a comment here.

dissect/cstruct/tools/stubify.py
```diff
@@ -0,0 +1,105 @@
...
+def setup_logger(verbosity: int) -> None:
+    if verbosity == 0:
+        log.setLevel(level=logging.WARNING)
+    elif verbosity == 1:
+        log.setLevel(level=logging.INFO)
+    elif verbosity > 1:
+        log.setLevel(level=logging.DEBUG)
+
+
+def main():
```
Schamper:
```suggestion
def main() -> None:
```

dissect/cstruct/tools/stubify.py
```diff
@@ -0,0 +1,105 @@
...
+        iterator = [file_path]
+
+    for file in iterator:
+        if file.is_file() and ".py" in file.suffixes:
+            stub = stubify_file(file, file_path)
+            if not stub:
+                continue
+
+            with file.with_suffix(".pyi").open("wt") as output_file:
+                log.info(f"Writing stub of file {file} to {output_file}")
```
Schamper: `%s` logging.

dissect/cstruct/tools/stubify.py
```diff
@@ -0,0 +1,105 @@
...
+    setup_logger(args.verbose)
+
+    file_path: Path = args.path
+
+    iterator = file_path.rglob("*.py")
+    if file_path.is_file():
+        iterator = [file_path]
+
+    for file in iterator:
+        if file.is_file() and ".py" in file.suffixes:
```
Schamper: Anywhere in the suffixes or _the_ suffix?

dissect/cstruct/types/packed.py
```diff
@@ -63,3 +63,8 @@ def _write(cls, stream: BinaryIO, data: Packed) -> int:
     @classmethod
     def _write_array(cls, stream: BinaryIO, data: list[Packed]) -> int:
         return stream.write(_struct(cls.cs.endian, f"{len(data)}{cls.packchar}").pack(*data))
+
+    @classmethod
+    def to_stub(cls, name: str):
```
Schamper:
```suggestion
    def to_stub(cls, name: str) -> None:
```

dissect/cstruct/types/packed.py
```diff
@@ -63,3 +63,8 @@ def _write(cls, stream: BinaryIO, data: Packed) -> int:
...
+        types = ", ".join([x.__name__ for x in cls.__bases__])
+        return f"{name}= type[{types}]"
```
Schamper:
```suggestion
        return f"{name} = type[{types}]"
```

dissect/cstruct/types/structure.py
```diff
@@ -28,6 +28,9 @@ def __repr__(self) -> str:
         bits_str = f" : {self.bits}" if self.bits else ""
         return f"<Field {self.name} {self.type.__name__}{bits_str}>"
 
+    def type_stub(self):
```
Schamper:
```suggestion
    def _type_stub(self) -> str:
```

dissect/cstruct/types/structure.py
```diff
@@ -363,6 +366,24 @@ def commit(cls) -> None:
         for key, value in classdict.items():
             setattr(cls, key, value)
 
+    def to_stub(cls, name: str = ""):
```
Schamper:
```suggestion
    def to_stub(cls, name: str = "") -> str:
```

dissect/cstruct/types/structure.py
```diff
@@ -363,6 +366,24 @@ def commit(cls) -> None:
...
+        with io.StringIO() as data:
```
Schamper: What's the additional value of using the context manager here?

dissect/cstruct/cstruct.py
```diff
@@ -406,6 +408,42 @@ def _make_union(
     ) -> type[Structure]:
         return self._make_struct(name, fields, align=align, anonymous=anonymous, base=Union)
 
+    def to_stub(self, name: str = "", ignore_type_defs: list[str] | None = None):
```
Schamper:
```suggestion
    def to_stub(self, name: str = "", ignore_type_defs: list[str] | None = None) -> str:
```

dissect/cstruct/cstruct.py
Schamper, line 1: I think the two added methods added in this file might be better served to just be functions in `stubify.py`, unless you can think of a good reason why they should be methods instead?

dissect/cstruct/types/base.py
```diff
@@ -181,6 +181,15 @@ def _write_0(cls, stream: BinaryIO, array: list[BaseType]) -> int:
         """
         return cls._write_array(stream, array + [cls()])
 
+    def _class_stub(cls) -> str:
+        return f"class {cls.__name__}({cls.__base__.__name__}):\n"
+
+    def _type_stub(cls, name: str = "") -> str:
+        return f"{name}: {cls.__name__}"
+
+    def to_stub(cls, name: str) -> str:
```
Schamper: Maybe name this differently? Also I think this might as well be a private method. Maybe `_to_type_stub`? `to_stub` might be confusing because it also makes me think of the compiler.

Schamper (changes requested):

dissect/cstruct/cstruct.py
Miauwkeru: Not specificially, i believe i used more internal parts in a previous iteration, which why I put it here. 
And i liked the `c_struct.to_stub()` call itself

dissect/cstruct/types/structure.py
Miauwkeru: At that point i didn't want to close it manually, I'll change it to be consistent

dissect/cstruct/types/base.py
Miauwkeru: i renamed it to `to_type_stub` as it is a function that gets called externally.

Schamper: I don't agree on the "externally" part, I'd classify this usage as internal and would never expect anyone to ever call this externally. This is not part of any expected usage API.

dissect/cstruct/tools/stubify.py
```diff
@@ -0,0 +1,153 @@
...
+        log.setLevel(level=logging.INFO)
+    elif verbosity > 1:
+        log.setLevel(level=logging.DEBUG)
+
+
+def main() -> None:
+    description = """
+        Create stub files for cstruct definitions.
+    
+        These stub files are in a `.pyi` format and provides `type` information to cstruct definitions.
```
Schamper: Why is `type` monospaced? Also this would need to be double back ticks anyway, if you're going to include this on the dissect.cstruct doc page.

dissect/cstruct/tools/stubify.py
```diff
@@ -0,0 +1,153 @@
...
+        This in turn gives a developer insight into the elements inside the definition and 
+        parameter completion when dealing with cstructs.
```
Schamper:
```suggestion
```

dissect/cstruct/tools/stubify.py
```diff
@@ -0,0 +1,153 @@
...
+
+    buffer = io.StringIO()
+    indentation = ""
+    if name:
+        buffer.write(f"class {name}(cstruct):\n")
+        indentation = " " * 4
+
+    prev_offset = buffer.tell()
+    for const, value in c_structure.consts.items():
+        buffer.write(indent(f"{const}: {type(value).__name__}=...\n", prefix=indentation))
```
Schamper:
```suggestion
        buffer.write(indent(f"{const}: {type(value).__name__} = ...\n", prefix=indentation))
```

dissect/cstruct/tools/stubify.py
Schamper, line 1: Can you add some unit tests for this file? All the coverage warnings make it difficult to review.

dissect/cstruct/types/enum.py
```diff
@@ -84,6 +85,19 @@ def _write_0(cls, stream: BinaryIO, array: list[BaseType]) -> int:
...
+
+    def to_type_stub(cls, name: str = "") -> str:
+        output = ""
+        with io.StringIO() as buf:
+            buf.write(cls._class_stub())
+            for key in cls.__members__.keys():
+                buf.write(f"    {key} = ...\n")
+            output = buf.getvalue()
+
+        return output
```
Schamper:
```suggestion
        result = [cls._class_stub()]
        for key in cls.__members__.keys():
            result.append(f"    {key} = ...")
        return "\n".join(result)
```
Should also get rid of the newline in `_class_stub`. Might be confusing for custom type implementations that you're responsible for ending with a newline.

pyproject.toml
```diff
@@ -31,6 +31,9 @@ homepage = "https://dissect.tools"
 documentation = "https://docs.dissect.tools/en/latest/projects/dissect.cstruct"
 repository = "https://github.com/fox-it/dissect.cstruct"
 
+[project.scripts]
+stubify = "dissect.cstruct.tools.stubify:main"
```
Schamper:
```suggestion
cstruct-stubify = "dissect.cstruct.tools.stubify:main"
```

tests/test_stub.py
```diff
@@ -0,0 +1,132 @@
...
+            "Test",
+            """
+            class Test(Structure):
+                class __anonymous_0__(Union):
+                    a: WcharArray
+                    b: CharArray
+                    def __init__(self, a: WcharArray=..., b: CharArray=...): ...
+                a: WcharArray
+                b: CharArray
+                def __init__(self, __anonymous_0__: __anonymous_0__=...): ...
```
Schamper: Anonymous structures should get folded in, I don't think there's any use in them being typed?

dissect/cstruct/types/base.py
```diff
@@ -181,6 +181,15 @@ def _write_0(cls, stream: BinaryIO, array: list[BaseType]) -> int:
         """
         return cls._write_array(stream, array + [cls()])
 
+    def _class_stub(cls) -> str:
+        return f"class {cls.__name__}({cls.__base__.__name__}):\n"
```
Schamper:
```suggestion
        return f"class {cls.__name__}({cls.__base__.__name__}):"
```

dissect/cstruct/types/enum.py
```diff
@@ -84,6 +85,19 @@ def _write_0(cls, stream: BinaryIO, array: list[BaseType]) -> int:
         data = [entry.value if isinstance(entry, Enum) else entry for entry in array]
         return cls._write_array(stream, data + [cls.type()])
 
+    def _class_stub(cls) -> str:
+        return f"class {cls.__name__}({cls.__base__.__name__}, {cls.type.__name__}):\n"
```
Schamper:
```suggestion
        return f"class {cls.__name__}({cls.__base__.__name__}, {cls.type.__name__}):"
```

dissect/cstruct/types/structure.py
```diff
@@ -363,6 +366,27 @@ def commit(cls) -> None:
...
+        for field in cls.fields.values():
+            type_info = field.type_stub()
+            buffer.write(indent(f"{type_info}\n", prefix=" " * 4))
+
+        call = ", ".join(call_args)
+        buffer.write(indent(f"def __init__({call}): ...\n", prefix=" " * 4))
+
+        output = buffer.getvalue()
+        buffer.close()
+        return output
```
Schamper:
```suggestion
        result = [f"class {cls.__name__}({cls.__base__.__name__}):"]

        args = ["self"]
        for field in cls.fields.values():
            if isinstance(field.type, StructureMetaType) and not field.type.__anonymous__:
                result.append(indent(field.type.to_type_stub(), prefix=" " * 4))

            result.append(f"    {field._type_stub()}")
            args.append(f"{field._type_stub()} = ...")

        result.append(indent(f"def __init__({', '.join(args)}): ...", prefix=" " * 4))
        return "\n".join(result)
```

Schamper (changes requested):

Miauwkeru:
I added some more tests, and noticed there was still some issues with import stuff from a file. As all signatures/names dissapear once the stub file is there. I attempted to fix this by adding more checks in `stubify_file`. This seems good enough for now, but I think it might be better to use ast in the future. And maybe look if we can use `mypy` to generate the stubs with hooks.

Schamper:
I have made a lot of suggested changes with https://github.com/fox-it/dissect.cstruct/pull/72/commits/7c93f22dd11f51a4e035ee5e84693aa10c34b986

pyproject.toml
```diff
@@ -31,12 +31,16 @@ homepage = "https://dissect.tools"
...
+[project.scripts]
+cstruct-stubgen = "dissect.cstruct.tools.stubgen:main"
+
 [tool.ruff]
 line-length = 120
 required-version = ">=0.9.0"
 
 [tool.ruff.format]
 docstring-code-format = true
+exclude = ["tests/data/*"]
```
Schamper:
```suggestion
```

pyproject.toml
```diff
@@ -74,6 +78,7 @@ select = [
   "RUF",
 ]
 ignore = ["E203", "B904", "UP024", "ANN002", "ANN003", "ANN204", "ANN401", "SIM105", "TRY003"]
+exclude = ["tests/data/*"]
```
Schamper:
```suggestion
```

Schamper (changes requested):

dissect/cstruct/tools/stubgen.py
```diff
@@ -0,0 +1,208 @@
...
+            continue
+
+        if issubclass(typedef, types.Enum):
+            stub = generate_enum_stub(typedef, cs_prefix=cs_prefix, module_prefix=module_prefix)
+        elif issubclass(typedef, types.Structure):
+            stub = generate_structure_stub(typedef, cs_prefix=cs_prefix, module_prefix=module_prefix)
+        elif issubclass(typedef, types.BaseType):
+            stub = generate_generic_stub(typedef, cs_prefix=cs_prefix, module_prefix=module_prefix)
+        elif isinstance(typedef, str):
+            stub = f"{name}: TypeAlias = {typedef}"
```
Miauwkeru, line 96: I don't think this code will ever get reached. As the TypeAlias where there for the uint8 and those kinda types. These are now defined in cstruct proper no?

dissect/cstruct/types/structure.py
```diff
@@ -258,21 +310,21 @@ def _read(cls, stream: BinaryIO, context: dict[str, Any] | None = None) -> Struc
...
-                    value = field_type(bit_buffer.read(field_type.type, field.bits))
+                if isinstance(field.type, EnumMetaType):
+                    value = field.type(bit_buffer.read(field.type.type, field.bits))
                 else:
-                    value = bit_buffer.read(field_type, field.bits)
+                    value = bit_buffer.read(field.type, field.bits)
 
-                if field.name:
-                    result[field.name] = value
+                if field._name:
```
Miauwkeru: isn't field._name always defined as it either uses `name` or `type.__name__`?

dissect/cstruct/types/structure.py
```diff
@@ -258,21 +310,21 @@ def _read(cls, stream: BinaryIO, context: dict[str, Any] | None = None) -> Struc
...
 
-            value = field_type._read(stream, result)
+            value = field.type._read(stream, result)
 
-            if field.name:
-                sizes[field.name] = stream.tell() - offset
-                result[field.name] = value
+            if field._name:
+                sizes[field._name] = stream.tell() - offset
+                result[field._name] = value
```
Miauwkeru: same here

dissect/cstruct/tools/stubgen.py
Schamper: Could technically be manually added I think.

dissect/cstruct/tools/stubgen.py
```diff
@@ -0,0 +1,218 @@
...
+    defined_names = set()
+
+    # Then typedefs
+    for name, typedef in cs.typedefs.items():
+        if name in empty_cs.typedefs:
+            continue
+
+        if typedef.__name__ in defined_names:
+            # Create an alias to the type if we have already seen it before.
+            stub = f"{name}: {typedef.__name__}"
```
Schamper: This is not an alias now, is it? This makes it an instance of the other one.

Also, can you add a unit test for this behaviour?

dissect/cstruct/types/structure.py
Schamper: I don't know why this `if` is even here in the first place 😄. But I was too scared to remove it. I think without all the other changes, `field.name` was _also_ always defined.

tests/test_types_union.py
```diff
@@ -270,7 +270,7 @@ def test_union_definition_nested(cs: cstruct, compiled: bool) -> None:
     assert isinstance(obj.c.a, UnionProxy)
     assert len(obj.c.a) == 8
     assert bytes(obj.c.a) == b"holybeef"
-    assert repr(obj.c.a) == "<a a=0x796c6f68 b=0x66656562>"
+    assert repr(obj.c.a) == "<__anonymous_0__ a=0x796c6f68 b=0x66656562>"
```
Schamper, line 273: I'm not sure if this is better or worse. I think with this change to the field names and anonymous structure names, we're technically more correct, but it's less legible when using e.g. `dumpstruct`.

Maybe we can patch the class and qual-name when we learn of the field name?

tests/test_basic.py
```diff
@@ -439,12 +446,12 @@ def test_size_and_aligment(cs: cstruct) -> None:
 
 def test_dynamic_substruct_size(cs: cstruct) -> None:
     cdef = """
-    struct {
+    typedef struct {
```
Schamper: I'm not a C expert but I think prior to this change, `sub` was the variable name and not a structure name, so this is now technically more correct? But is it "nicer"? Or should we still allow this kind of definition, if not just for the sake of backwards compatibility?

Schamper (changes requested):

tests/test_basic.py
Miauwkeru: From some digging that I did, cause I am also not an expert on C. Yes, previously `sub` was a variable declaration of an anonymous structure.
Now it is a type alias of an anonymous structure. So now it is more correct.

As for whether we should allow this kind of definition, probably not if we want to completely match the c spec. I think this would warrant a larger discussion regarding this.
But regardless, I wouldn't immediately remove the ability to declare the "name" this way. If we want to go that route, we should first properly put a deprecation notice that you won't be able to create a structure of a specific type like this in the future.

tests/test_types_union.py
Miauwkeru: hmm, good question. It does feel a bit hacky tbh. But if we do go that route, would it be an idea to use some kind of convention for anonymous structures like these? For example `__a` or `__a__` in this case? 
We would need to document that part ofc, but it is at least less visual noise.

tests/test_basic.py
Schamper: I made a suggested fix for this with https://github.com/fox-it/dissect.cstruct/commit/83505dd40fce387ade688b73549b9735360ec6d9

Miauwkeru: Those change LGTM, anything else we need to do for this tool?

Schamper: I'll have a look over once more.

tests/test_types_union.py
Schamper: I looked at this briefly but let's just keep this behavior for now and look at it later. It's easier for the stub generation. We can make it prettier later.

tests/test_basic.py
Schamper: Probably good 👍

Schamper (approved):

PR #88, opened 2024-05-05
Ignore tabs in expression

### LINKED ISSUE #87: Tabs in constant expression should be treated like spaces
For example:

```c
#define A<tab>(B + C)<tab>// comment
```

currently errors when parsing, even though tabs are legal whitespace separators.

d3dave:
Fixes #87

Constant expression can contain tabs as whitespace just like spaces. This PR tests for this and fixes the expression parser to ignore the tabs just like spaces.


dissect/cstruct/expression.py
```diff
@@ -18,7 +18,7 @@ def __init__(self, expression: str):
         self.pos = 0
         self.tokens = []
 
-    def equal(self, token: str, expected: Union[str, str[str]]) -> bool:
+    def equal(self, token: str, expected: Union[str, set[str]]) -> bool:
```
Schamper:
```suggestion
    def equal(self, token: str, expected: str | set[str]) -> bool:
```
May as well use the new syntax here too.

Schamper (changes requested):

d3dave:
@DissectBot agree

Schamper (changes requested):
If you fix the linter this is good to go :+1:

Schamper (approved):

PR #91, opened 2024-05-24
Fix enum membership checks

Schamper:


pyrco (approved):

dissect/cstruct/types/enum.py
```diff
@@ -56,6 +57,13 @@ def __getitem__(cls, name: str | int) -> Enum | Array:
 
     __len__ = MetaType.__len__
 
+    if not PY_312:
+        # Backport __contains__ from CPython 3.12
+        def __contains__(cls, value: Any) -> bool:
+            if isinstance(value, cls):
+                return True
+            return value in cls._value2member_map_
```
cecinestpasunepipe, line 65:
```suggestion
        def __contains__(cls, value: Any) -> bool:
            if isinstance(value, cls):
                return True
            return hasattr(cls, value)
```

value2member map is not what the code in ntfs seems to expect...

cecinestpasunepipe (dismissed):

dissect/cstruct/types/enum.py
Schamper: The code in ntfs was wrong.

cecinestpasunepipe: or there must be a mechanism to get the member by string?

PR #93, opened 2024-06-03
fix offsets in generate fields

Miauwkeru:
- Unpack enum when generating fields.
- Align bit fields better

### DIFF
--- a/dissect/cstruct/compiler.py
+++ b/dissect/cstruct/compiler.py
@@ -198,8 +198,11 @@ class _ReadSourceGenerator:
                     prev_was_bits = True
 
                 if bits_remaining == 0 or prev_bits_type != field_type:
-                    bits_remaining = (size * 8) - field.bits
-                    bits_rollover = True
+                    if bits_remaining == field.bits:
+                        bits_remaining = 0
+                    else:
+                        bits_remaining = (size * 8) - field.bits
+                        bits_rollover = True
 
                 yield from flush()
                 yield from align_to_field(field)
--- a/dissect/cstruct/compiler.py
+++ b/dissect/cstruct/compiler.py
@@ -109,6 +109,7 @@ class _ReadSourceGenerator:
         preamble = """
         r = {}
         s = {}
+        o = stream.tell()
         """
 
         if any(field.bits for field in self.fields):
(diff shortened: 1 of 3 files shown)


dissect/cstruct/compiler.py
```diff
@@ -190,10 +195,10 @@ def align_to_field(field: Field) -> Iterator[str]:
...
             elif field.bits:
                 if not prev_was_bits:
-                    prev_bits_type = field.type
+                    prev_bits_type = field_type
                     prev_was_bits = True
 
-                if bits_remaining == 0 or prev_bits_type != field.type:
+                if bits_remaining == 0 or prev_bits_type != field_type:
                     bits_remaining = (size * 8) - field.bits
                     bits_rollover = True
```
Miauwkeru, line 203: Note:
This will still leave space in a structure. 
so with the following example:
```python
struct data {
  uint64 a: 4,
  uint16 b: 4,
};

data(a=4, b=5).dumps() == b"\x04\x00\x00\x00\x00\x00\x00\x00\x05\x00"
```
this is done for both reading and writing, which might result in unwanted stuff

tests/test_compiler.py
```diff
@@ -324,3 +324,82 @@ def test_generate_bits_read(cs: cstruct, TestEnum: type[Enum]) -> None:
...
+
+    _s = stream.tell()
+    r["c"] = _3._read(stream, context=r)
+    s["c"] = stream.tell() - _s
+    """
+
+    assert output.strip() == dedent(expected).strip()
+
+
+# TODO: the rest of the compiler
```
pyrco:
```suggestion
```

tests/test_compiler.py
```diff
@@ -324,3 +324,82 @@ def test_generate_bits_read(cs: cstruct, TestEnum: type[Enum]) -> None:
 
 
 # TODO: the rest of the compiler
```
pyrco:
```suggestion
```

tests/test_compiler.py
```diff
@@ -324,3 +324,82 @@ def test_generate_bits_read(cs: cstruct, TestEnum: type[Enum]) -> None:
...
+    r["c"] = _3._read(stream, context=r)
+    s["c"] = stream.tell() - _s
+    """
+
+    assert output.strip() == dedent(expected).strip()
+
+
+# TODO: the rest of the compiler
+@pytest.mark.parametrize("other_type", ["int8", "uint64"])
+def test_generate_fields_dynamic_after_bitfield_reverse(cs: cstruct, TestEnum: Enum, other_type: str) -> None:
```
pyrco:
```suggestion
def test_generate_fields_dynamic_before_bitfield(cs: cstruct, TestEnum: Enum, other_type: str) -> None:
```

pyrco (changes requested):

pyrco (approved):

PR #101, opened 2024-09-23
Fix enum and flag repr in struct reprs

### LINKED ISSUE #100: cstruct v4 prints enums as integers instead of the enum itself
In `dissect.cstruct` v4 the enum types are now printed as hex integers instead of the enum type itself.
This affects printing a structure but also in dumpstruct. In my opinion printing only the integer loses some contextual information that is useful when dumping structures.

The `examples/mirai.py` in the cstruct repository shows the difference in after and before:

dissect.cstruct==4.1:

image

dissect.cstruct==3.9:

image

Notice the difference in `AttackOption type=`

Schamper:
Fixes #100

### DIFF
--- a/dissect/cstruct/types/structure.py
+++ b/dissect/cstruct/types/structure.py
@@ -381,10 +381,15 @@ class Structure(BaseType, metaclass=StructureMetaType):
         return getattr(self, item)
 
     def __repr__(self) -> str:
-        values = [
-            f"{k}={hex(self[k]) if (issubclass(f.type, int) and not issubclass(f.type, (Pointer, Enum))) else repr(self[k])}"
-            for k, f in self.__class__.fields.items()
-        ]
+        values = []
+        for k, f in self.__class__.fields.items():
+            value = self[k]
+            if issubclass(f.type, int) and not issubclass(f.type, (Pointer, Enum)):
+                value = hex(value)
+            else:
+                value = repr(value)
+            values.append(f"{k}={value}")
+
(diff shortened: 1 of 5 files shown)


dissect/cstruct/types/structure.py
```diff
@@ -380,10 +381,15 @@ def __getitem__(self, item: str) -> Any:
         return getattr(self, item)
 
     def __repr__(self) -> str:
-        values = [
-            f"{k}={hex(self[k]) if (issubclass(f.type, int) and not issubclass(f.type, Pointer)) else repr(self[k])}"
-            for k, f in self.__class__.fields.items()
-        ]
+        values = []
+        for k, f in self.__class__.fields.items():
```
Miauwkeru: no single character variable names pls

Miauwkeru (dismissed):

yunzheng:
please also fix the same behavior in `dumpstruct()` https://github.com/fox-it/dissect.cstruct/blob/046dfa51f0989c7f55cf6d7dd00ba71f3ec5a115/dissect/cstruct/utils.py#L163

Example output from a `dissect.cobaltstrike` test:
image

yunzheng (changes requested):
please also fix the behaviour in `dumpstruct()`

Schamper:
@yunzheng should be fixed now too.

yunzheng (approved):
Confirmed that this is now fixed in both repr and dumpstruct output.

PR #104, opened 2024-10-22
Fix default initialization for various types

### LINKED ISSUE #102: Errors when initialising a struct with partial keywords or arguments
When initialising a struct with partial keywords or arguments the other fields become `None` resulting in exceptions, for example when printing the instance of the struct.

When initialising without any arguments it works and all values are 0 as expected. But when you only initialise some of the fields it breaks, example:

```python

C_DEF = """"
struct FooBar {
    uint32  foo;
    uint32  bar;
};
"""

cs = cstruct().load(C_DEF)

print(cs.FooBar())  # works as expected: <FooBar foo=0x0 bar=0x0>
print(cs.FooBar(1, 1)) # works as expected: <FooBar foo=0x1 bar=0x1>

print(cs.FooBar(1)) # breaks, bar=None
print(cs.FooBar(bar=1)) # breaks, foo=None
```

The exception:

```
$ python3 kwargs_init.py 
<FooBar foo=0x0 bar=0x0>
<FooBar foo=0x1 bar=0x1>
Traceback (most recent call last):
  File "/workspaces/dissect.cstruct/examples/kwargs_init.py", line 14, in <module>
    print(cs.FooBar(1)) # breaks, bar=None
    ^^^^^^^^^^^^^^^^^^^
  File "/workspaces/dissect.cstruct/dissect/cstruct/types/structure.py", line 388, in __repr__
    value = hex(value)
            ^^^^^^^^^^
TypeError: 'NoneType' object cannot be interpreted as an integer
```

Schamper:
Fixes #102 and the bug that occurred in https://github.com/fox-it/dissect.hypervisor/pull/31.


dissect/cstruct/types/base.py
```diff
@@ -225,6 +225,11 @@ class ArrayMetaType(MetaType):
     num_entries: int | Expression | None
     null_terminated: bool
 
+    def default(cls) -> BaseType:
+        return type.__call__(
+            cls, [cls.type.default() for _ in range(cls.num_entries if isinstance(cls.num_entries, int) else 0)]
```
twiggler:
```suggestion
            cls, [cls.type.default()] * (cls.num_entries if isinstance(cls.num_entries, int) else 0)
```

dissect/cstruct/types/pointer.py
```diff
@@ -16,7 +16,7 @@ class Pointer(int, BaseType):
     _context: dict[str, Any]
     _value: BaseType
 
-    def __new__(cls, value: int, stream: BinaryIO, context: dict[str, Any] = None) -> Pointer:
+    def __new__(cls, value: int, stream: BinaryIO | None, context: dict[str, Any] = None) -> Pointer:
```
twiggler:
```suggestion
    def __new__(cls, value: int, stream: BinaryIO | None = None, context: dict[str, Any] = None) -> Pointer:
```

(curious)

If `context` can be `None`, should the type not reflect this, i.e. `dict[str, Any] | None`?
This would propagate to the `_context` field, and possibly further.

Perhaps an empty dictionary is safer than `None`, although possibly memory inefficient?

dissect/cstruct/types/pointer.py
```diff
@@ -16,7 +16,7 @@ class Pointer(int, BaseType):
...
         obj = super().__new__(cls, value)
         obj._stream = stream
```
twiggler, line 21: If stream can be `None`, then the `_stream` field should be `BinaryIO | None` I think

dissect/cstruct/types/pointer.py
```diff
@@ -74,7 +78,7 @@ def _write(cls, stream: BinaryIO, data: int) -> int:
         return cls.cs.pointer._write(stream, data)
 
     def dereference(self) -> Any:
-        if self == 0:
+        if self == 0 or self._stream is None:
```
twiggler, line 81: Is it  a valid state to have `self == 0 and self._stream is not None`, or 
`self != 0 and self._stream is None`?

If not, perhaps we should assert this invariant earlier, for instance during construction of the pointer?

dissect/cstruct/types/structure.py
```diff
@@ -271,7 +271,7 @@ def _read(cls, stream: BinaryIO, context: dict[str, Any] = None) -> Structure:
             # Align the stream
             stream.seek(-stream.tell() & (cls.alignment - 1), io.SEEK_CUR)
 
-        obj = cls(**result)
+        obj = type.__call__(cls, **result)
```
twiggler, line 277: Can you add a comment why this is necessary?

dissect/cstruct/types/pointer.py
Schamper: Yes, those are valid states. 

`self == 0 and self._stream is not None`: we have a source stream, but the read/parsed pointer value is 0.
`self != 0 and self._stream is None`: We initialize an "empty" pointer instance with a value, but we don't have a source stream. This is a valid state for writing.

dissect/cstruct/types/structure.py
```diff
@@ -397,11 +397,24 @@ class UnionMetaType(StructureMetaType):
...
+            # We don't support user initialization of dynamic unions yet
+            if cls.dynamic:
+                raise NotImplementedError("Initializing a dynamic union is not yet supported")
+
+            # User (partial) initialization, rebuild the union
+            # First user-provided field is the one used to rebuild the union
+            arg_fields = [field.name for _, field in zip(args, cls.__fields__)]
+            kwarg_fields = [name for name in kwargs if name in cls.lookup]
+            arg_fields.extend(kwarg_fields)
+            obj._rebuild(next(iter(arg_fields)))
```
twiggler: I think `arg_fields` can be an empty list at this point? 
If so, this might throw some exception, which is might be difficult for user to understand.

dissect/cstruct/types/structure.py
```diff
@@ -397,11 +397,24 @@ class UnionMetaType(StructureMetaType):
...
+
+        # Calling with non-stream args or kwargs means we are initializing with values
+        if (args and not (len(args) == 1 and (_is_readable_type(args[0]) or _is_buffer_type(args[0])))) or kwargs:
+            # We don't support user initialization of dynamic unions yet
+            if cls.dynamic:
+                raise NotImplementedError("Initializing a dynamic union is not yet supported")
+
+            # User (partial) initialization, rebuild the union
+            # First user-provided field is the one used to rebuild the union
+            arg_fields = [field.name for _, field in zip(args, cls.__fields__)]
```
twiggler: I like that there are no branches in this code.

However, it seems a bit wasteful to create lists of these args and kwargs, when we are only interested in taking the first one. This could be a performance-critical section?

dissect/cstruct/types/structure.py
```diff
@@ -696,3 +741,83 @@ def __hash__(self):
...
+
+def _generate_structure__init__(fields: list[Field]) -> FunctionType:
+    """Generates an ``__init__`` method for a structure with the specified fields.
+
+    Args:
+        fields: List of field names.
+    """
+    field_names = [field.name for field in fields]
+
+    template: FunctionType = _make_structure__init__(len(field_names))
```
twiggler, line 774: To understand this function you need to simultaenously read _make_structure__init__, since we are modifying the code generated there. The naming also suggests that they are strongly related.

Perhaps making `_make_structure__init` an inner function of `_generate_structure__init__` is clearer?

Schamper: We lose the caching benefit in that case, as the cache would get re-initialized on every execution of `_generate_structure__init__`.

dissect/cstruct/types/structure.py
Schamper: I think we crash before we get here in the scenario where `arg_fields` could be empty, but changed it regardless.

dissect/cstruct/types/structure.py
Schamper: This is luckily not a hot path, but I changed to to generators anyway. Also fixes the other comment you had.

dissect/cstruct/types/structure.py
twiggler: Ah yes. 

`lru_cache` doesn't support explicitly referencing the cache so that we can make the cache a global variable. 

Ok, nothing to be done.

twiggler (approved):
Nice!

Lked the "concept checks" like `is_readable_type`.

PR #111, opened 2024-11-29
Add support  for newlines inside "NAME" token

sMezaOrellana:
The following file is parsable by the gcc & clang compiler.

```c
#define __SOCKADDR_COMMON_SIZE (sizeof(unsigned short))
typedef uint32_t in_addr_t;

#define __SOCKADDR_COMMON(sa_prefix) sa_family_t sa_prefix##family

typedef unsigned short sa_family_t;
struct in_addr {
  in_addr_t s_addr;
};

typedef uint16_t in_port_t;

struct sockaddr {
  sa_family_t sa_family;

  char sa_data[14]; /* Address data.  */
};

struct sockaddr_in {
  sa_family_t sin_family;
  in_port_t sin_port;      /* Port number.  */
  struct in_addr sin_addr; /* Internet address.  */

  /* Pad to size of `struct sockaddr'.  */
  unsigned char sin_zero[sizeof(struct sockaddr) - __SOCKADDR_COMMON_SIZE -
                        sizeof(in_port_t) - sizeof(struct in_addr)];
};
```

however this part of the code was giving issues when parsing with cstruct

```c 
  /* Pad to size of `struct sockaddr'.  */
  unsigned char sin_zero[sizeof(struct sockaddr) - __SOCKADDR_COMMON_SIZE -
                        sizeof(in_port_t) - sizeof(struct in_addr)];
};
```

I modified the "NAME" token regex to allow for newline character. This however means that the new newline needs to be handled by the ExpressionTokenizer. Which was also added.

A test case was also added  to test the functionality.


Horofic (dismissed):

Schamper (changes requested):
Can you fix the conflict? Sorry for the huge delay by the way, this was dropped from my radar.

tests/test_types_structure.py
```diff
@@ -801,6 +801,39 @@ def __init__(self, _0 = None, _1 = None, _2 = None, _3 = None, _4 = None):
...
+    obj.magic = "test"
+    obj.wmagic = "test"
+    obj.a = 0x01
+    obj.b = 0x0203
+    obj.c = 0x04050607
+    obj.string = b"lalala"
+    obj.wstring = "test"
+
+    assert obj.dumps() == buf
+
```
Schamper:
```suggestion


```
I'm not sure how this passes the linter, but I think this should have two newlines.

I think line 279 is wrong too, can you maybe also add a newline there?

Schamper (changes requested):

Schamper (approved):

PR #114, opened 2025-01-29
Set types as attributes on the cstruct object

### LINKED ISSUE #113: Eagerly resolve types and set as attributes on the cstruct object
Currently all type accesses (`cs.uint8`) go through `__getattr__`, and if it's not a constant, `cstruct.resolve()`. This is very slow. We should look into if there are any reasons why we can't resolve types eagerly and set them as instance attributes on the `cstruct` object. This would make type accesses a lot faster.

Random thoughts:
- There may be a specific reason we do dynamic resolutions with `.resolve()`, but how big is that use case? For example, I suppose it would allow changing the typedef of a field and have that dynamically resolve at read, but this is already not possible with compiled structures (where we resolve types at compile time). To be fair, that use case is currently still possible if you opt-out your structure for compilation.
- Maybe we should only do it for specific types of things, for example `enum`, `flag`, `struct` and constant definitions. And then we can catch `typedef` with the existing `__getattr__`. That way we do allow for dynamic type changes, but not for things that are supposed to be static. Since dynamic typedef'ing would be an advanced topic anyway, the intended "workaround" for that would be to use `typedef` and use the `typedef`'d name instead of the `struct` name. "Performance oriented code" could then use the raw `struct` and `enum` names for a faster access time (properly written code utilising cstruct already does this to make the loop count in `.resolve()` as low as possible).

Schamper:
Closes #113 

Some micro benchmarks:

```python

cdef = """
#define X 512

enum MyEnum {
    A,
    B,
    C
};

struct test {
    uint32 a;
};
"""
cs = cstruct()
cs.load(cdef)
```

Before:
```python
In: %timeit getattr(t.cs, "X")
58.1 ns ± 1.04 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In: %timeit getattr(t.cs, "MyEnum")
227 ns ± 0.912 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In: %timeit getattr(t.cs, "test")
219 ns ± 1.47 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

After:
```python
In: %timeit getattr(t.cs, "X")
31.2 ns ± 0.0331 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In: %timeit getattr(t.cs, "MyEnum")
33.1 ns ± 0.0884 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In: %timeit getattr(t.cs, "test")
31.3 ns ± 0.0571 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

### DIFF
--- a/dissect/cstruct/cstruct.py
+++ b/dissect/cstruct/cstruct.py
@@ -220,7 +220,7 @@ class cstruct:
         return name
 
     def _add_attr(self, name: str, value: Any, replace: bool = False) -> None:
-        if not replace and (name in self.__dict__ and self.__dict__[name] != value):
+        if not replace and ((existing := self.__dict__.get(name)) is not None and existing != value):
             raise ValueError(f"Attribute already exists: {name}")
         setattr(self, name, value)
 
--- a/dissect/cstruct/cstruct.py
+++ b/dissect/cstruct/cstruct.py
@@ -52,9 +52,12 @@ class cstruct:
 
         self.consts = {}
         self.lookups = {}
+        self.types = {}
+        self.typedefs = {}
         self.includes = []
+
         # fmt: off
-        self.typedefs = {
+        initial_types = {
             # Internal types
             "int8": self._make_packed_type("int8", "b", int),
             "uint8": self._make_packed_type("uint8", "B", int),
@@ -97,6 +100,21 @@ class cstruct:
             "long long": "int64",
             "signed long long": "int64",
             "unsigned long long": "uint64",
+
+            # Other convenience types
+            "u1": "uint8",
+            "u2": "uint16",
+            "u4": "uint32",
+            "u8": "uint64",
+            "u16": "uint128",
+            "__u8": "uint8",
+            "__u16": "uint16",
+            "__u32": "uint32",
+            "__u64": "uint64",
+            "uchar": "uint8",
+            "ushort": "uint16",
+            "uint": "uint32",
+            "ulong": "uint32",
 
             # Windows types
             "BYTE": "uint8",
@@ -165,23 +183,11 @@ class cstruct:
             "_DWORD": "uint32",
             "_QWORD": "uint64",
             "_OWORD": "uint128",
-
-            # Other convenience types
-            "u1": "uint8",
-            "u2": "uint16",
-            "u4": "uint32",
-            "u8": "uint64",
-            "u16": "uint128",
-            "__u8": "uint8",
-            "__u16": "uint16",
-            "__u32": "uint32",
-            "__u64": "uint64",
-            "uchar": "uint8",
-            "ushort": "uint16",
-            "uint": "uint32",
-            "ulong": "uint32",
         }
         # fmt: on
+
+        for name, type_ in initial_types.items():
+            self.add_type(name, type_)
 
         pointer = pointer or ("uint64" if sys.maxsize > 2**32 else "uint32")
         self.pointer: type[BaseType] = self.resolve(pointer)
@@ -197,35 +203,69 @@ class cstruct:
             pass
 
         try:
+            return self.types[attr]
+        except KeyError:
+            pass
+
+        try:
             return self.resolve(self.typedefs[attr])
         except KeyError:
             pass
 
-        raise AttributeError(f"Invalid attribute: {attr}")
+        return super().__getattribute__(attr)
 
     def _next_anonymous(self) -> str:
         name = f"__anonymous_{self._anonymous_count}__"
         self._anonymous_count += 1
         return name
 
+    def _add_attr(self, name: str, value: Any, replace: bool = False) -> None:
+        if not replace and (name in self.__dict__ and self.__dict__[name] != value):
+            raise ValueError(f"Attribute already exists: {name}")
+        setattr(self, name, value)
+
     def add_type(self, name: str, type_: type[BaseType] | str, replace: bool = False) -> None:
         """Add a type or type reference.
 
         Only use this method when creating type aliases or adding already bound types.
+        All types will be resolved to their actual type objects prior to being added.
+        Use :func:`add_typedef` to add type references.
 
         Args:
             name: Name of the type to be added.
             type_: The type to be added. Can be a str reference to another type or a compatible type class.
+                   If a str is given, it will be resolved to the actual type object.
 
         Raises:
             ValueError: If the type already exists.
         """
+        typeobj = self.resolve(type_)
+        if not replace and (name in self.types and self.types[name] != typeobj):
+            raise ValueError(f"Duplicate type: {name}")
+
+        self.types[name] = typeobj
+        self._add_attr(name, typeobj, replace=replace)
+
+    addtype = add_type
+
+    def add_typedef(self, name: str, type_: str, replace: bool = False) -> None:
+        """Add a type reference.
+
+        Use this method to add type references to this cstruct instance. These are type names that can be
+        dynamically resolved at a later stage. Use :func:`add_type` to add actual type objects.
+
+        Args:
+            name: Name of the type to be added.
+            type_: The type reference to be added.
+            replace: Whether to replace the type if it already exists.
+        """
+        if not isinstance(type_, str):
+            raise TypeError("Type reference must be a string")
+
         if not replace and (name in self.typedefs and self.resolve(self.typedefs[name]) != self.resolve(type_)):
             raise ValueError(f"Duplicate type: {name}")
 
         self.typedefs[name] = type_
-
-    addtype = add_type
 
     def add_custom_type(
         self, name: str, type_: type[BaseType], size: int | None = None, alignment: int | None = None, **kwargs
@@ -243,6 +283,16 @@ class cstruct:
             **kwargs: Additional attributes to add to the type.
         """
         self.add_type(name, self._make_type(name, (type_,), size, alignment=alignment, attrs=kwargs))
+
+    def add_const(self, name: str, value: Any) -> None:
+        """Add a constant value.
+
+        Args:
+            name: Name of the constant to be added.
+            value: The value of the constant.
+        """
+        self.consts[name] = value
+        self._add_attr(name, value, replace=True)
 
     def load(self, definition: str, deftype: int | None = None, **kwargs) -> cstruct:
         """Parse structures from the given definitions using the given definition type.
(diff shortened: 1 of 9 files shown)


Miauwkeru:
Some initial remarks with these changes, some projects fail to build:
- `dissect.executable` fails because it cannot find the `Elf_type`. My assumption is because at https://github.com/fox-it/dissect.executable/blob/01df4c2e66227e7f084b98c5c2b97f0b40a8d4f5/dissect/executable/elf/c_elf.py#L396 the types dict doesn't get copied.
-  `dissect.shellitem` fails because the code checks whether something is inside the `typedefs` which now moved to `types`: https://github.com/fox-it/dissect.shellitem/blob/16bf53da8d5a31bee57e6788a54b04e761d55108/dissect/shellitem/lnk/lnk.py#L76
- `dissect.target`: jumplist also fails because of the shellitem issue.

dissect/cstruct/cstruct.py
```diff
@@ -196,37 +202,71 @@ def __getattr__(self, attr: str) -> Any:
...
-        raise AttributeError(f"Invalid attribute: {attr}")
+        return super().__getattribute__(attr)
 
     def _next_anonymous(self) -> str:
         name = f"__anonymous_{self._anonymous_count}__"
         self._anonymous_count += 1
         return name
 
+    def _add_attr(self, name: str, value: Any, replace: bool = False) -> None:
+        if not replace and (name in self.__dict__ and self.__dict__[name] != value):
```
Miauwkeru:
```suggestion
        if not replace and (self.__dict__.get(name) != value):
```
Or is the overhead of the function call higher?

dissect/cstruct/tools/stubgen.py
```diff
@@ -79,32 +79,43 @@ def generate_cstruct_stub(cs: cstruct, module_prefix: str = "", cls_name: str =
```
Miauwkeru, line 79: There is an issue on L174/L185 where it now generates structures multiple times due to looking into `typedefs` instead of `types`

Miauwkeru (changes requested):

dissect/cstruct/tools/stubgen.py
Miauwkeru: From what I have seen with the following example:
```c
struct SYSTEMTIME {
    WORD    wYear;
    WORD    wMonth;
    WORD    wDayOfWeek;
    WORD    wDay;
    WORD    wHour;
    WORD    wMinute;
    WORD    wSecond;
    WORD    wMilliseconds;
};

struct TimeZoneInformation {
    LONG        Bias;
    wchar       StandardName[32];
    SYSTEMTIME  StandardDate;
    LONG        StandardBias;
    wchar       DaylightName[32];
    SYSTEMTIME  DaylightDate;
    LONG        DaylightBias;
};
```
It will generate a stub file like this:
```python
class _c_struct(__cs__.cstruct):
    class SYSTEMTIME(__cs__.Structure):
        wYear: _c_struct.uint16
        wMonth: _c_struct.uint16
        wDayOfWeek: _c_struct.uint16
        wDay: _c_struct.uint16
        wHour: _c_struct.uint16
        wMinute: _c_struct.uint16
        wSecond: _c_struct.uint16
        wMilliseconds: _c_struct.uint16
        @overload
        def __init__(
            self,
            wYear: _c_struct.uint16 | None = ...,
            wMonth: _c_struct.uint16 | None = ...,
            wDayOfWeek: _c_struct.uint16 | None = ...,
            wDay: _c_struct.uint16 | None = ...,
            wHour: _c_struct.uint16 | None = ...,
            wMinute: _c_struct.uint16 | None = ...,
            wSecond: _c_struct.uint16 | None = ...,
            wMilliseconds: _c_struct.uint16 | None = ...,
        ): ...
        @overload
        def __init__(self, fh: bytes | memoryview | bytearray | BinaryIO, /): ...

    class TimeZoneInformation(__cs__.Structure):
        Bias: _c_struct.int32
        StandardName: __cs__.WcharArray
        class SYSTEMTIME(__cs__.Structure):
            wYear: _c_struct.uint16
            wMonth: _c_struct.uint16
            wDayOfWeek: _c_struct.uint16
            wDay: _c_struct.uint16
            wHour: _c_struct.uint16
            wMinute: _c_struct.uint16
            wSecond: _c_struct.uint16
            wMilliseconds: _c_struct.uint16
            @overload
            def __init__(
                self,
                wYear: _c_struct.uint16 | None = ...,
                wMonth: _c_struct.uint16 | None = ...,
                wDayOfWeek: _c_struct.uint16 | None = ...,
                wDay: _c_struct.uint16 | None = ...,
                wHour: _c_struct.uint16 | None = ...,
                wMinute: _c_struct.uint16 | None = ...,
                wSecond: _c_struct.uint16 | None = ...,
                wMilliseconds: _c_struct.uint16 | None = ...,
            ): ...
            @overload
            def __init__(self, fh: bytes | memoryview | bytearray | BinaryIO, /): ...

        StandardDate: SYSTEMTIME
        StandardBias: _c_struct.int32
        DaylightName: __cs__.WcharArray
        class SYSTEMTIME(__cs__.Structure):
            wYear: _c_struct.uint16
            wMonth: _c_struct.uint16
            wDayOfWeek: _c_struct.uint16
            wDay: _c_struct.uint16
            wHour: _c_struct.uint16
            wMinute: _c_struct.uint16
            wSecond: _c_struct.uint16
            wMilliseconds: _c_struct.uint16
            @overload
            def __init__(
                self,
                wYear: _c_struct.uint16 | None = ...,
                wMonth: _c_struct.uint16 | None = ...,
                wDayOfWeek: _c_struct.uint16 | None = ...,
                wDay: _c_struct.uint16 | None = ...,
                wHour: _c_struct.uint16 | None = ...,
                wMinute: _c_struct.uint16 | None = ...,
                wSecond: _c_struct.uint16 | None = ...,
                wMilliseconds: _c_struct.uint16 | None = ...,
            ): ...
            @overload
            def __init__(self, fh: bytes | memoryview | bytearray | BinaryIO, /): ...

        DaylightDate: SYSTEMTIME
        DaylightBias: _c_struct.int32
        @overload
        def __init__(
            self,
            Bias: _c_struct.int32 | None = ...,
            StandardName: __cs__.WcharArray | None = ...,
            StandardDate: SYSTEMTIME | None = ...,
            StandardBias: _c_struct.int32 | None = ...,
            DaylightName: __cs__.WcharArray | None = ...,
            DaylightDate: SYSTEMTIME | None = ...,
            DaylightBias: _c_struct.int32 | None = ...,
        ): ...
        @overload
        def __init__(self, fh: bytes | memoryview | bytearray | BinaryIO, /): ...
```
Here `SYSTEMTIME` gets generated multiple times within the structure, as it can't see that `SYSTEMTIME` was already defined, due to looking into `typedefs` instead of `types`

Schamper:
https://github.com/fox-it/dissect.executable/pull/39 will resolve the `dissect.executable` one. As for shellitem, as previously discussed I'd like to propose to inline that into `dissect.target` and resolve it there.

dissect/cstruct/tools/stubgen.py
```diff
@@ -171,7 +168,7 @@ def generate_structure_stub(
         while issubclass(nested_type, types.BaseArray):
             nested_type = nested_type.type
 
-        if issubclass(nested_type, types.Structure) and type_name not in structure.cs.typedefs:
+        if issubclass(nested_type, types.Structure) and type_name not in structure.cs.types:
```
Miauwkeru:
```suggestion
        if issubclass(nested_type, types.Structure) and nested_type.__name__ not in structure.cs.types:
```
I noticed that those fields that use a struct e.g. kept generating a nested struct while it was available elsewhere. This was because those types where of `<type _IMAGE_DATA_DIRECTORY[16]>` instead of `<type _IMAGE_DATA_DIRECTORY>`

Tho that was not related to this pr.

dissect/cstruct/cstruct.py
```diff
@@ -231,22 +233,34 @@ def _next_anonymous(self) -> str:
         self._anonymous_count += 1
         return name
 
+    def _add_attr(self, name: str, value: Any) -> None:
+        # Names that collide with the cstruct class (e.g. a struct named ``load``) are not set as attributes
+        # to avoid breaking the instance. They remain accessible through ``resolve`` and the type dicts.
+        if name in _RESERVED_NAMES:
+            return
+        setattr(self, name, value)
+
```
Miauwkeru, line 242: I assume it never / rarely happens, but should we take into account that someone can shadow a struct with a define or anonymous enum?

Miauwkeru (changes requested):

dissect/cstruct/tools/stubgen.py
Schamper: I don't understand?

dissect/cstruct/cstruct.py
Schamper: I don't understand?

Miauwkeru: Right now, if a `define` or a `struct` or members of an anonymous enum use the same name, they keep overwriting each other right?

e.g. in the following situation

```c
struct test {
  uint8 data;
};

#define test
flag {
  test,
}
```
The cstruct.test will now point to the flag `test` if i understand the code correctly. So it overwrites the attributes on the cstruct instance.
The same occurs with the `consts` where `test` also points to the same flag.

Schamper: Just like real C?

dissect/cstruct/tools/stubgen.py
Miauwkeru: let's use the following example

```c
typedef struct _IMAGE_DATA_DIRECTORY {
    ULONG       VirtualAddress;
    ULONG       Size;
} IMAGE_DATA_DIRECTORY, *PIMAGE_DATA_DIRECTORY;

typedef struct _IMAGE_OPTIONAL_HEADER {
   ...
   IMAGE_DATA_DIRECTORY  DataDirectory[16];
} IMAGE_OPTIONAL_HEADER32
```
When generating the stub for `_IMAGE_OPTIONAL_HEADER` the previous behavior used the class name of the field type for `DataDirectory` which is `_IMAGE_DATA_DIRECTORY[16]`. 

That type (`_IMAGE_DATA_DIRECTORY[16]`) does not exist in `structure.cs.types`  so it generates a nested stub definition for `_IMAGE_DATA_DIRECTORY`.

So I propose to use the `nested_type.__name__`, which would be `_IMAGE_DATA_DIRECTORY` and is available in `structure.cs.types`. To avoid those duplicate nested stubs.

As for:
> Tho that was not related to this pr.

This isn't an issue specific to this pr, but as you are changing the line anyways it might be useful to take it along.

dissect/cstruct/cstruct.py
Miauwkeru: ah, yes, I didn't check the real C part.

Miauwkeru (dismissed):

Schamper:
had to fix the linter

Miauwkeru (approved):

PR #122, opened 2025-05-19
Save space saving sizes of dynamic fields only

twiggler:
Currently,  actual sizes of all fields are stored inside the `_sizes` private member. This can cause high memory usage in some applications. 

This PR only stores the sizes of dynamic fields. Since most fields are static, we significantly reduce memory usage, utilities like `dumpstruct` still work as intended. 

Alternative solution considered: Remove `_sizes` completely. However, no alternative way has been found to reconstruct the number of bytes read. This would result in loss of the pretty colors in dump_struct, for example

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,53 @@
-# cstruct version 4.6
-Release date: tbc
+# Changelog
 
-## Library
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
+
+## [Unreleased]
+
+### Changed
+
 - Optimize storage of field sizes.
-- Rename `_sizes` property of `Structure` to `__sizes__`. 
+- Rename `_sizes` property of `Structure` to `__sizes__`.
 - Rename `_values` property of `Structure` to `__values__`.
+
+## [4.5] - 20-05-2025
+
+### Added 
+
+- Introduce experimental tool `cstruct-stubgen` to generate type stubs for cstruct definitions.
+
+### Fixed
+
+- Generated classes are now hashable.
+- Suppress spurious `TypeError: Dynamic size` errors when using cstruct interactively.
+
+## [4.4] - 03-10-2025
+
+### Fixed
+
+- Resolve documentation warnings.
+
+### Changed
+
+- Use the Ruff linter.
+
+## [4.3] - 11-18-2024
+
+### Fixed
+
+- All cstruct types are now correctly default-initialized using the `__default__` member.
+
+## [4.2] - 10-10-2024
+
+### Fixed
+
+- The string representation of enums now outputs the name of the constants. 
+
+## [4.1] - 10-09-2024
+
+### Fixed
+
+- Declaring an array of a nested struct type now works as intended.
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,7 @@
+# cstruct version 4.6
+Release date: tbc
+
+## Library
+- Optimize storage of field sizes.
+- Rename `_sizes` property of `Structure` to `__sizes__`. 
+- Rename `_values` property of `Structure` to `__values__`.
--- a/dissect/cstruct/types/structure.py
+++ b/dissect/cstruct/types/structure.py
@@ -79,8 +79,8 @@ class StructureMetaType(MetaType):
             return type.__call__(cls, *args, **kwargs)
         if not args and not kwargs:
             obj = type.__call__(cls)
-            object.__setattr__(obj, "_values", {})
-            object.__setattr__(obj, "_sizes", {})
+            object.__setattr__(obj, "__values__", {})
+            object.__setattr__(obj, "__dynamic_sizes__", {})
             return obj
 
         return super().__call__(*args, **kwargs)
@@ -282,8 +282,8 @@ class StructureMetaType(MetaType):
         # Using type.__call__ directly calls the __init__ method of the class
         # This is faster than calling cls() and bypasses the metaclass __call__ method
         obj = type.__call__(cls, **result)
-        obj._sizes = sizes
-        obj._values = result
+        obj.__dynamic_sizes__ = sizes
+        obj.__values__ = result
         return obj
 
     def _read_0(cls, stream: BinaryIO, context: dict[str, Any] | None = None) -> list[Self]:  # type: ignore
@@ -378,8 +378,8 @@ class StructureMetaType(MetaType):
 class Structure(BaseType, metaclass=StructureMetaType):
     """Base class for cstruct structure type classes."""
 
-    _values: dict[str, Any]
-    _sizes: dict[str, int]
+    __values__: dict[str, Any]
+    __dynamic_sizes__: dict[str, int]
 
     def __len__(self) -> int:
         return len(self.dumps())
@@ -403,12 +403,12 @@ class Structure(BaseType, metaclass=StructureMetaType):
         return f"<{self.__class__.__name__} {' '.join(values)}>"
 
     @cached_property
-    def sizes(self) -> None:
+    def __sizes__(self) -> dict[str, int | None]:
         """Return the sizes of the fields in this structure."""
         sizes = {}
         for field in self.__class__.__fields__:
             if field.type.dynamic:
-                sizes[field._name] = self._sizes.get(field._name, None)
+                sizes[field._name] = self.__dynamic_sizes__.get(field._name, None)
             else:
                 sizes[field._name] = field.type.size
 
@@ -510,8 +510,8 @@ class UnionMetaType(StructureMetaType):
         # It also makes it easier to differentiate between user-initialization of the class
         # and initialization from a stream read
         obj: Union = type.__call__(cls, **result)
-        object.__setattr__(obj, "_values", result)
-        object.__setattr__(obj, "_sizes", sizes)
+        object.__setattr__(obj, "__values__", result)
+        object.__setattr__(obj, "__dynamic_sizes__", sizes)
         object.__setattr__(obj, "_buf", buf)
 
         if cls.size is not None:
@@ -593,8 +593,8 @@ class Union(Structure, metaclass=UnionMetaType):
     def _update(self) -> None:
         result, sizes = self.__class__._read_fields(io.BytesIO(self._buf))
         self.__dict__.update(result)
-        object.__setattr__(self, "_values", result)
-        object.__setattr__(self, "_sizes", sizes)
+        object.__setattr__(self, "__values__", result)
+        object.__setattr__(self, "__dynamic_sizes__", sizes)
 
     def _proxify(self) -> None:
         def _proxy_structure(value: Structure) -> None:
--- a/dissect/cstruct/types/structure.py
+++ b/dissect/cstruct/types/structure.py
@@ -412,12 +412,12 @@ class Structure(BaseType, metaclass=StructureMetaType):
         return f"<{self.__class__.__name__} {' '.join(values)}>"
 
     @property
-    def __sizes__(self) -> Mapping[str, int | None]:
-        return ChainMap(self.__class__.__static_sizes__, self.__dynamic_sizes__)
-
-    @property
     def __values__(self) -> MutableMapping[str, Any]:
         return StructureValuesProxy(self)
+
+    @property
+    def __sizes__(self) -> Mapping[str, int | None]:
+        return ChainMap(self.__class__.__static_sizes__, self.__dynamic_sizes__)
 
 
 class StructureValuesProxy(MutableMapping):
@@ -433,9 +433,8 @@ class StructureValuesProxy(MutableMapping):
 
     def __setitem__(self, key: str, value: Any) -> None:
         if key in self:
-            setattr(self._struct, key, value)
-        else:
-            raise KeyError(key)
+            return setattr(self._struct, key, value)
+        raise KeyError(key)
 
     def __contains__(self, key: str) -> bool:
         return key in self._struct.__class__.fields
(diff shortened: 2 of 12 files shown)


dissect/cstruct/utils.py
```diff
@@ -173,6 +168,10 @@ def _dumpstruct(
                 value = value.replace("\n", f"\n{' ' * (len(field._name) + 4)}")
 
         if color:
+            foreground, background = colors[ci % len(colors)]
+            size = structure._sizes[field._name] if field.type.dynamic else field.type.size
```
twiggler, line 172: Reviewer: This logic could be moved into a `sizes` property on the cstruct type.

dissect/cstruct/types/structure.py
```diff
@@ -403,6 +402,18 @@ def __repr__(self) -> str:
 
         return f"<{self.__class__.__name__} {' '.join(values)}>"
 
+    @cached_property
+    def sizes(self) -> None:
```
Schamper: We can't really add a `sizes` property since it will clash with structures that have a field names `sizes`. Which is why almost everything related to these classes is either `_` or `__` 😄.

Schamper (changes requested):

dissect/cstruct/types/structure.py
twiggler: Ah yes, a similar mistake as the  `default` method IIRC. 

I renamed to `__field_sizes__`

JSCU-CNI: Wouldn't it make more sense to keep using `_sizes` instead of `__field_sizes__`?

dissect/cstruct/types/structure.py
```diff
@@ -401,6 +407,46 @@ def __repr__(self) -> str:
...
+        return StructureValuesProxy(self)
+
+
+class StructureValuesProxy(MutableMapping):
+    """A proxy for the values of fields of a Structure."""
+
+    def __init__(self, struct: Structure):
+        self._struct: Structure = struct
+        # Unfortunately, we need to support code that adds attributes dynamically :/
+        self._fields: set[str] = set(iter(self._struct.__class__.fields))
```
twiggler: Alternatively, we could use a ChainMap here over field names and extra fields. This would be more memory efficient, but slightly slower when iterating (since chainmap needs to preserve key order).

dissect/cstruct/types/structure.py
```diff
@@ -401,6 +407,46 @@ def __repr__(self) -> str:
...
+    def __values__(self) -> MutableMapping[str, Any]:
+        return StructureValuesProxy(self)
+
+
+class StructureValuesProxy(MutableMapping):
+    """A proxy for the values of fields of a Structure."""
+
+    def __init__(self, struct: Structure):
+        self._struct: Structure = struct
+        # Unfortunately, we need to support code that adds attributes dynamically :/
```
Schamper: We do? To a structure object? Where does such code live?

dissect/cstruct/types/structure.py
```diff
@@ -401,6 +407,46 @@ def __repr__(self) -> str:
...
+        return key in self._fields
+
+    def __iter__(self) -> Iterator[str]:
+        return iter(self._fields)
+
+    def __len__(self) -> int:
+        return len(self._fields)
+
+    def __repr__(self) -> str:
+        return repr(self._struct)
```
Schamper: Shouldn't the `repr` be more like a dictionary?

dissect/cstruct/types/structure.py
Schamper: What do you mean extra fields? You can't dynamically add fields to an already parsed structure.

dissect/cstruct/types/structure.py
```diff
@@ -401,6 +407,46 @@ def __repr__(self) -> str:
...
+class StructureValuesProxy(MutableMapping):
+    """A proxy for the values of fields of a Structure."""
+
+    def __init__(self, struct: Structure):
+        self._struct: Structure = struct
+        # Unfortunately, we need to support code that adds attributes dynamically :/
+        self._fields: set[str] = set(iter(self._struct.__class__.fields))
+
+    def __getitem__(self, key: str) -> Any:
+        return getattr(self._struct, key)
```
Schamper:
```suggestion
        if key in self._fields:
            return getattr(self._struct, key)
        raise KeyError(key)
```
?

dissect/cstruct/types/structure.py
```diff
@@ -401,6 +407,46 @@ def __repr__(self) -> str:
...
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        setattr(self._struct, key, value)
+        self._fields.add(key)
```
Schamper:
```suggestion
        if key in self._fields:
            setattr(self._struct, key, value)
        raise KeyError(key)
```
?

dissect/cstruct/types/structure.py
```diff
@@ -401,6 +407,46 @@ def __repr__(self) -> str:
...
+
+    def __delitem__(self, key: str) -> None:
+        del self._struct[key]
+        self._fields.discard(key)
```
Schamper: Don't think this should be allowed?

Schamper (changes requested):

dissect/cstruct/types/structure.py
twiggler: For example [here](https://github.com/fox-it/dissect.etl/blob/3219d88bb6c6668fedb40a5394519c3cef54ee89/dissect/etl/headers/event.py#L55)

Since  ETL has been programmed against an older version of cstruct, I asked MiauwKeru  and  he confirmed that adding attributes should be possible, even if the structure has no corresponding field.

Moreover, we don´t check when setting an attribute to see if it has a corresponding field, so I gathered that this is expected behavior.

Personally, I think adding new attributes this way is is a bad idea. 

Should I remove this capability? 
And perhaps add a check to `setattr` of structure that we don´t create attributes without corresponding fields?

dissect/cstruct/types/structure.py
twiggler: See above, although "extra fields" should have been "extra attributes"

dissect/cstruct/types/structure.py
twiggler: Yeah it seems like a bad idea, but we don´t guard against deleting fields from structures so I thought it should be possible.

Shall I disable?

dissect/cstruct/types/structure.py
```diff
@@ -401,6 +411,49 @@ def __repr__(self) -> str:
 
         return f"<{self.__class__.__name__} {' '.join(values)}>"
 
+    @property
+    def __sizes__(self) -> Mapping[str, int | None]:
+        return ChainMap(self.__class__.__static_sizes__, self.__dynamic_sizes__)
+
+    @property
+    def __values__(self) -> MutableMapping[str, Any]:
+        return StructureValuesProxy(self)
```
Schamper:
```suggestion
    @property
    def __values__(self) -> MutableMapping[str, Any]:
        return StructureValuesProxy(self)

    @property
    def __sizes__(self) -> Mapping[str, int | None]:
        return ChainMap(self.__class__.__static_sizes__, self.__dynamic_sizes__)
```
Just for consistency sake.

dissect/cstruct/types/structure.py
```diff
@@ -401,6 +411,49 @@ def __repr__(self) -> str:
...
+
+    def __iter__(self) -> Iterator[str]:
+        return iter(self._struct.__class__.fields)
+
+    def __len__(self) -> int:
+        return len(self._struct.__class__.fields)
+
+    def __repr__(self) -> str:
+        items = (f"{key!r}: {value!r}" for (key, value) in self.items())
+        return "{" + ", ".join(items) + "}"
```
Schamper:
```suggestion
        return repr(dict(self))
```
Would this work?

dissect/cstruct/types/structure.py
```diff
@@ -401,6 +411,49 @@ def __repr__(self) -> str:
...
+    def __getitem__(self, key: str) -> Any:
+        if key in self:
+            return getattr(self._struct, key)
+        raise KeyError(key)
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        if key in self:
+            setattr(self._struct, key, value)
+        else:
+            raise KeyError(key)
```
Schamper:
```suggestion
        raise KeyError(key)
```

dissect/cstruct/types/structure.py
```diff
@@ -401,6 +411,49 @@ def __repr__(self) -> str:
...
+        self._struct: Structure = struct
+
+    def __getitem__(self, key: str) -> Any:
+        if key in self:
+            return getattr(self._struct, key)
+        raise KeyError(key)
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        if key in self:
+            setattr(self._struct, key, value)
```
Schamper:
```suggestion
            return setattr(self._struct, key, value)
```

tests/test_compiler.py
```diff
@@ -337,6 +318,7 @@ def test_generate_fields_dynamic_after_bitfield(cs: cstruct, TestEnum: Enum, oth
         Field("b", _type, 4),
         Field("c", cs.char["size"], offset=3),
     ]
+    fields[3].type.dynamic = True   # cs.char["size"] is a hack so patch dynamic
```
Schamper: What do you mean with it being a hack?

Schamper (changes requested):

dissect/cstruct/types/structure.py
Schamper: @JSCU-CNI we've decided to rename them to `__values__` and `__sizes__`. This is more in line with the rest of the "internal attribute" naming and formalizes it a bit more.

tests/test_compiler.py
twiggler: AFAIK   `Field("c", cs.char["size"], offset=3)` does not set dynamic is True. 

Maybe mock is a better term

Schamper: Looks like faulty usage of this. `_make_array` expects an integer size or an `Expression` here (from the syntax parser). On the one hand I believe this should be allowed usage [since it's possible to add fields yourself from code](https://github.com/fox-it/dissect.cstruct/blob/e990b5301ffa6c739cdbd954baeb70d3a970a362/dissect/cstruct/types/structure.py#L354-L359). I think it can be solved by simply wrapping the string in an `Expression` in `_make_array`.

Or we don't allow this, in which case we should turn this "mock" more proper by making it `cs.char[Expression("size")]`.

twiggler: When we creating fields programmatically I prefer the more explicit `cs.char[Expression("size")]`. 

However, the Expression constructor also required a cstruct instance. Because the cstruct instance is only required during evaluation, I made it a parameter of `Expression::evaluate`

dissect/cstruct/types/structure.py
twiggler: meh :/

dissect/cstruct/types/structure.py
twiggler: Tested fine, it is more elegant

tests/test_compiler.py
```diff
@@ -377,6 +358,7 @@ def test_generate_fields_dynamic_before_bitfield(cs: cstruct, TestEnum: Enum, ot
         Field("b", TestEnum, 4),
         Field("c", cs.char["size"], offset=3),
     ]
+    fields[3].type.dynamic = True   # cs.char["size"] is a hack so patch dynamic
```
Schamper: Fix here too?

Schamper (changes requested):

Schamper (dismissed):

JSCU-CNI:
Could this change be documented, e.g. using a `CHANGELOG` file entry?

CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# cstruct version 4.6
+Release date: tbc
+
+## Library
+- Optimize storage of field sizes.
+- Rename `_sizes` property of `Structure` to `__sizes__`. 
+- Rename `_values` property of `Structure` to `__values__`.
```
Schamper:
```suggestion
# Changelog

All notable changes to this project will be documented in this file.

The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).

## [Unreleased]

### Changed

- Optimize storage of field sizes
- Rename `_sizes` property of `Structure` to `__sizes__`.
- Rename `_values` property of `Structure` to `__values__`

[unreleased]: https://github.com/fox-it/dissect.cstruct/compare/4.5...HEAD
```
I looked online for some examples of nice changelog formats and stumbled upon this, maybe we can try it? What do you think.

I'm thinking the comparison URLs are a bit superfluous with what's already available in https://github.com/fox-it/dissect.cstruct/releases, so maybe we should leave that out. Also maybe consider adding the previous changelogs in here too?

Schamper (changes requested):

CHANGELOG.md
twiggler: > What do you think.

I don´t have a strong opinion about the format. The format you propose looks fine. With regard to the contents, it think it is important that entries are written with the user in mind.

I do wonder how the changelog compares and contrasts with the github releases functionality in *general* (i.e. https://github.com/fox-it/dissect.cstruct/releases). It seems they share the same purpose? Maybe we can automate keeping them in sync.

> I'm thinking the comparison URLs are a bit superfluous with what's already available in https://github.com/fox-it/dissect.cstruct/releases, so maybe we should leave that out. 

I think they look ugly, in favor of dropping them.

> Also maybe consider adding the previous changelogs in here too?

I can add the changes starting from 4.0. Version four seems to be a complete rewrite, and I don´t have enough context to write a meaningful changelog beyond. Personally I wouldn´t bother though.

Schamper (approved):

PR #134, opened 2025-07-18
Allow cstruct class initialization with definition

### LINKED ISSUE #133: Allow cstruct class initialization directly with a definition
It's quite cumbersome to initialize a `cstruct` object using:

```python
cs = cstruct().load(c_def)
```

It would be much easier to initialize it with a definition:

```python
cs = cstruct(c_def)
```

And this is also easier to remember as there is one less step to lookup in the documentation.

yunzheng:
This quality of life improvement allows the user to directly initialize a `cstruct` object using a c-definition.

This change also enforces that `endian` or `pointer` arguments (and future arguments) must be passed using keyword arguments only. It seems that no code base used positional arguments for this before (checked using github codesearch), but this is a breaking change if someone did.

Tests are included.

resolves #133

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Optimize storage of field sizes.
 - Rename `_sizes` property of `Structure` to `__sizes__`.
 - Rename `_values` property of `Structure` to `__values__`.
-- Added `load` argument to `cstruct` class, other arguments to `cstruct` are now keyword only.
+- Added `load` argument to `cstruct` class, allowing direct initialization with a definition (i.e. `cstruct(cdef)` instead of `cstruct().load(cdef)`. Other arguments to `cstruct` are now keyword only.
 
 ## [4.5] - 20-05-2025
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Optimize storage of field sizes.
 - Rename `_sizes` property of `Structure` to `__sizes__`.
 - Rename `_values` property of `Structure` to `__values__`.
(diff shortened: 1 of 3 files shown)


yunzheng:
Note that it doesn't have a `compiled=` kwarg or other kwargs like `load()` does.

I don't think it's necessary but it would be possible to add.

Schamper:
> This change also enforces that `endian` or `pointer` arguments (and future arguments) must be passed using keyword arguments only. It seems that no code base used positional arguments for this before (checked using github codesearch), but this is a breaking change if someone did.

Can you add it to the `CHANGELOG.md`?

Schamper:
> Note that it doesn't have a `compiled=` kwarg or other kwargs like `load()` does.
> 
> I don't think it's necessary but it would be possible to add.

Don't think it's necessary either, if you require it (or another possible future/advanced argument for `load()`) you should use that method instead. This is a shortcut.

yunzheng:
> > This change also enforces that `endian` or `pointer` arguments (and future arguments) must be passed using keyword arguments only. It seems that no code base used positional arguments for this before (checked using github codesearch), but this is a breaking change if someone did.
> 
> Can you add it to the `CHANGELOG.md`?

Done, please check if it describes the change well enough.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Optimize storage of field sizes.
 - Rename `_sizes` property of `Structure` to `__sizes__`.
 - Rename `_values` property of `Structure` to `__values__`.
+- Added `load` argument to `cstruct` class, other arguments to `cstruct` are now keyword only.
```
Schamper:
```suggestion
- Added `load` argument to `cstruct` class, allowing direct initialization with a definition (i.e. `cstruct(cdef)` instead of `cstruct().load(cdef)`. Other arguments to `cstruct` are now keyword only.
```

Schamper (changes requested):

Schamper (approved):

PR #135, opened 2025-07-22
Add `#ifdef`/`#ifndef` parsing

Schamper:
Fixes #125

PR #137, opened 2025-07-31
Add support for `#undef`

Schamper:
As the title implies.

### DIFF
--- a/dissect/cstruct/parser.py
+++ b/dissect/cstruct/parser.py
@@ -54,7 +54,8 @@ class TokenParser(Parser):
     def _tokencollection() -> TokenCollection:
         TOK = TokenCollection()
         TOK.add(r"#\[(?P<values>[^\]]+)\](?=\s*)", "CONFIG_FLAG")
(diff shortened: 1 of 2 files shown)


dissect/cstruct/parser.py
```diff
@@ -54,7 +54,8 @@ def __init__(self, cs: cstruct, compiled: bool = True, align: bool = False):
     def _tokencollection() -> TokenCollection:
         TOK = TokenCollection()
         TOK.add(r"#\[(?P<values>[^\]]+)\](?=\s*)", "CONFIG_FLAG")
-        TOK.add(r"#define\s+(?P<name>[^\s]+)\s+(?P<value>[^\r\n]+)\s*", "DEFINE")
+        TOK.add(r"#define\s+(?P<name>[^\s]+)(?P<value>[^\r\n]*)", "DEFINE")
```
Miauwkeru, line 59: shouldn't the value of the "DEFINE" be stripped because of this? at least for python 3.9 if I understand the failing tests correctly. or something else goes wrong here

Schamper: #135 has this fix already so I'll just wait to merge that first.

Miauwkeru (approved):

Issue #143, opened 2026-02-11
Add helper to temporarily switch endianness


Schamper:
Sometimes we have to dynamically switch endianness for the lifetime of a parser (not the lifetime of the global cstruct instance). It would be nice to easily have the ability to switch the endianness for either a single struct parsing or of a "shallow" clone of the cstruct instance.

Maybe https://github.com/fox-it/dissect.cstruct/pull/140 is already "easy enough", unsure. Open to suggestions.

yunzheng:
Maybe a contextmanager would be nice, implemented like this after merging #140:

```python

class cstruct:
    ...
    @contextmanager
    def endianness(self, byte_order: str) -> Iterator[cstruct]:
        new = self.copy()
        new.endian = byte_order
        yield new
```

Example usage:

```python
from dissect.cstruct import cstruct

data = bytes.fromhex("AABBCCDD")

cdef = """
struct test {
    uint32  a;
};
"""

cs = cstruct(cdef, endian="<")

print(cs.test(data))

with cs.endianness(">") as big:
    print(big.test(data))
```

Outputs:

```shell
<test a=0xddccbbaa>
<test a=0xaabbccdd>
```

And another convenience method like `big_endian()` and `little_endian()` could also be added like this:

```python
    big_endian = partialmethod(endianness, byte_order=">")
    little_endian = partialmethod(endianness, byte_order="<")
```

Schamper:
I was thinking something in the direction of a context manager, that's a very nice approach!

Another idea I had could be `cs.test(data, endian="<")`. Although that would be a lot harder change to make.

PR #140, opened 2026-01-13
Add copy method to cstruct instance

Schamper:
Closes #79

PR #144, opened 2026-02-12
Add endian keyword argument to type reads

Schamper:
Partially solves https://github.com/fox-it/dissect.cstruct/issues/143, but I'd still like the context manager approach to be in there too. But that depends on https://github.com/fox-it/dissect.cstruct/pull/140.

Adds the ability to override the endianness on single type reads. So for example:

```python
In [1]: from dissect.cstruct import cstruct

In [2]: cs = cstruct(endian="<")

In [3]: hex(cs.uint32(b"\x01\x02\x03\x04"))
Out[3]: '0x4030201'

In [4]: hex(cs.uint32(b"\x01\x02\x03\x04", endian=">"))
Out[4]: '0x1020304'
```

This should work on any type, including structs, unions and pointers.

Because the function signature for type reading and writing changes because of this, I also added a check when adding a custom type to ensure it's signature is correct. If it's not, it will either throw a (descriptive) error or issue a warning. I also intentionally added a `**kwargs` to the function signature to allow for future expansion without breaking backwards compatibility again.

I'm not 100% if the behavior of pointers is as expected. I implemented it so that:
- If you don't override endianness, it'll take it from the cstruct instance (like everything else)
- If you override endianness, it'll be used for both the parsing of the pointer value and the dereferencing
- You can additionally override on `.dereference(endian=)`
- As with everything else, calling `.dumps()` will fall back to the cstruct instance endianness, unless overridden

### DIFF
--- a/dissect/cstruct/cstruct.py
+++ b/dissect/cstruct/cstruct.py
@@ -1,13 +1,15 @@
 from __future__ import annotations
 
 import ctypes as _ctypes
+import inspect
 import struct
 import sys
 import types
+import warnings
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, BinaryIO, TypeVar, cast
+from typing import TYPE_CHECKING, Any, BinaryIO, Literal, TypeVar, cast
 
-from dissect.cstruct.exceptions import ResolveError
+from dissect.cstruct.exceptions import Error, ResolveError
 from dissect.cstruct.expression import Expression
 from dissect.cstruct.parser import CStyleParser, TokenParser
 from dissect.cstruct.types import (
@@ -27,6 +29,7 @@ from dissect.cstruct.types import (
     Void,
     Wchar,
 )
+from dissect.cstruct.types.base import normalize_endianness
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
@@ -35,20 +38,23 @@ if TYPE_CHECKING:
 
 T = TypeVar("T", bound=BaseType)
 
+AllowedEndianness: TypeAlias = Literal["little", "big", "network", "<", ">", "!", "@", "="]
+Endianness: TypeAlias = Literal["<", ">", "!", "@", "="]
+
 
 class cstruct:
     """Main class of cstruct. All types are registered in here.
 
     Args:
-        endian: The endianness to use when parsing.
+        endian: The endianness to use when parsing (little, big, network, <, >, !, @ or =).
         pointer: The pointer type to use for pointers.
     """
 
     DEF_CSTYLE = 1
     DEF_LEGACY = 2
 
-    def __init__(self, load: str = "", *, endian: str = "<", pointer: str | None = None):
-        self.endian = endian
+    def __init__(self, load: str = "", *, endian: AllowedEndianness = "<", pointer: str | None = None):
+        self.endian = normalize_endianness(endian)
 
         self.consts = {}
         self.lookups = {}
@@ -242,6 +248,33 @@ class cstruct:
             alignment: The alignment of the type.
             **kwargs: Additional attributes to add to the type.
         """
+        # In cstruct 4.8 we changed the function signature of _read and _write
+        # Check if the function signature is compatible, and throw an error if not
+        for type_to_check in (type_, type_.ArrayType):
+            type_name = type_.__name__ + (f".{type_.ArrayType.__name__}" if type_to_check is type_.ArrayType else "")
+
+            for method in ("_read", "_read_array", "_read_0", "_write", "_write_array", "_write_0"):
+                if not hasattr(type_to_check, method):
+                    continue
+
+                signature = inspect.signature(getattr(type_to_check, method))
+
+                # We added a few keyword-only parameters to the function signature, but any custom type will
+                # continue to work fine as long as they accept **kwargs
+                if not any(param.kind == inspect.Parameter.VAR_KEYWORD for param in signature.parameters.values()):
+                    raise Error(
+                        f"Custom type {type_name} has an incompatible {method} method signature. "
+                        "Please refer to the changelog of dissect.cstruct 4.8 for more information."
+                    )
+
+                # Only warn if the method doesn't accept an endian parameter
+                if "endian" not in signature.parameters:
+                    warnings.warn(
+                        f"Custom type {type_name} is missing the 'endian' keyword-only parameter in its {method} method. "  # noqa: E501
+                        "Please refer to the changelog of dissect.cstruct 4.8 for more information.",
+                        stacklevel=2,
+                    )
+
         self.add_type(name, self._make_type(name, (type_,), size, alignment=alignment, attrs=kwargs))
 
     def load(self, definition: str, deftype: int | None = None, **kwargs) -> cstruct:
--- a/dissect/cstruct/cstruct.py
+++ b/dissect/cstruct/cstruct.py
@@ -2,11 +2,14 @@ from __future__ import annotations
 
 import copy
 import ctypes as _ctypes
+import inspect
 import struct
 import sys
 import types
+import warnings
+from contextlib import contextmanager
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, BinaryIO, TypeVar, cast
+from typing import TYPE_CHECKING, Any, BinaryIO, Literal, TypeVar, cast
 
 from dissect.cstruct.exception import ResolveError
 from dissect.cstruct.expression import Expression
@@ -26,9 +29,10 @@ from dissect.cstruct.types import (
     Void,
     Wchar,
 )
+from dissect.cstruct.types.base import normalize_endianness
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable
+    from collections.abc import Iterable, Iterator
     from typing import TypeAlias
 
     from dissect.cstruct.types import (
@@ -39,20 +43,23 @@ if TYPE_CHECKING:
 
 T = TypeVar("T", bound=BaseType)
 
+AllowedEndianness: TypeAlias = Literal["little", "big", "network", "<", ">", "!", "@", "="]
+Endianness: TypeAlias = Literal["<", ">", "!", "@", "="]
+
 
 class cstruct:
     """Main class of cstruct. All types are registered in here.
 
     Args:
-        endian: The endianness to use when parsing.
+        endian: The endianness to use when parsing (little, big, network, <, >, !, @ or =).
         pointer: The pointer type to use for pointers.
     """
 
     DEF_CSTYLE = 1
     DEF_LEGACY = 2
 
-    def __init__(self, load: str = "", *, endian: str = "<", pointer: str | None = None):
-        self.endian = endian
+    def __init__(self, load: str = "", *, endian: AllowedEndianness = "<", pointer: str | None = None):
+        self.endian = normalize_endianness(endian)
 
         self.consts: dict[str, int | str | bytes] = {}
         self.types: dict[str, type[BaseType]] = {}
@@ -279,6 +286,25 @@ class cstruct:
             alignment: The alignment of the type.
             **kwargs: Additional attributes to add to the type.
         """
+        # In cstruct 5.0 we changed the function signature of _read and _write
+        # Check if the function signature is compatible, and warn if not
+        for type_to_check in (type_, type_.ArrayType):
+            type_name = type_.__name__ + (f".{type_.ArrayType.__name__}" if type_to_check is type_.ArrayType else "")
+
+            for method in ("_read", "_read_array", "_read_0", "_write", "_write_array", "_write_0"):
+                if not hasattr(type_to_check, method):
+                    continue
+
+                signature = inspect.signature(getattr(type_to_check, method))
+
+                # Only warn if the method doesn't accept an endian parameter
+                if "endian" not in signature.parameters:
+                    warnings.warn(
+                        f"Custom type {type_name} is missing the 'endian' keyword-only parameter in its {method} method. "  # noqa: E501
+                        "Please refer to the changelog of dissect.cstruct 5.0 for more information.",
+                        stacklevel=2,
+                    )
+
         self.add_type(name, self._make_type(name, (type_,), size, alignment=alignment, attrs=kwargs))
 
     def add_const(self, name: str, value: Any) -> None:
@@ -416,6 +442,28 @@ class cstruct:
             A new cstruct instance with the same types and settings as this one.
         """
         return copy.copy(self)
+
+    @contextmanager
+    def endianness(self, endian: AllowedEndianness) -> Iterator[cstruct]:
+        """Context manager for temporarily changing the endianness of this cstruct instance."""
+        original = self.endian
+        self.endian = normalize_endianness(endian)
+        try:
+            yield self
+        finally:
+            self.endian = original
+
+    @contextmanager
+    def big_endian(self) -> Iterator[cstruct]:
+        """Context manager for temporarily changing the endianness to big endian."""
+        with self.endianness(">") as cs:
+            yield cs
+
+    @contextmanager
+    def little_endian(self) -> Iterator[cstruct]:
+        """Context manager for temporarily changing the endianness to little endian."""
+        with self.endianness("<") as cs:
+            yield cs
 
     def _make_type(
         self,
--- a/dissect/cstruct/types/base.py
+++ b/dissect/cstruct/types/base.py
@@ -12,7 +12,7 @@ if TYPE_CHECKING:
 
     from typing_extensions import Self
 
-    from dissect.cstruct.cstruct import cstruct
+    from dissect.cstruct.cstruct import AllowedEndianness, Endianness, cstruct
 
 
 EOF = -0xE0F  # Negative counts are illegal anyway, so abuse that for our EOF sentinel
@@ -41,14 +41,15 @@ class MetaType(type):
             stream = args[0]
 
             if _is_readable_type(stream):
-                return cls._read(stream)
+                endian = normalize_endianness(endian) if (endian := kwargs.get("endian")) is not None else cls.cs.endian
+                return cls._read(stream, endian=endian)
 
             if issubclass(cls, bytes) and isinstance(stream, bytes) and len(stream) == cls.size:
                 # Shortcut for char/bytes type
                 return type.__call__(cls, *args, **kwargs)
 
             if _is_buffer_type(stream):
-                return cls.reads(stream)
+                return cls.reads(stream, endian=kwargs.get("endian"))
 
         return type.__call__(cls, *args, **kwargs)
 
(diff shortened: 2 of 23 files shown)


Schamper:
I rebased on https://github.com/fox-it/dissect.cstruct/pull/140 and added the context manager.

dissect/cstruct/types/base.py
```diff
@@ -71,71 +72,91 @@ def __default__(cls) -> Self:  # type: ignore
         """Return the default value of this type."""
         return cls()
 
-    def reads(cls, data: bytes | memoryview | bytearray) -> Self:  # type: ignore
+    def reads(cls, data: bytes | memoryview | bytearray, *, endian: AllowedEndianness | None = None) -> Self:  # type: ignore
         """Parse the given data from a bytes-like object.
 
         Args:
             data: Bytes-like object to parse.
+            endian: The endianness to use when parsing. If not provided, the cstruct's default endianness will be used.
```
yunzheng: I found `the cstruct's default endianness` a bit ambiguous, as it sounds like it will use a global endianness defined in cstruct. And also in other docstrings you specify which options are available.

Maybe this is better:

```suggestion
            endian: The endianness to use when parsing (little, big, network, <, >, !, @ or =). When omitted, the cstruct instance's own endianness is used.
```

dissect/cstruct/types/base.py
```diff
@@ -71,71 +72,91 @@ def __default__(cls) -> Self:  # type: ignore
...
+        endian = normalize_endianness(endian) if endian is not None else cls.cs.endian
+
         out = BytesIO()
-        cls._write(out, value)
+        cls._write(out, value, endian=endian)
         return out.getvalue()
 
-    def _read(cls, stream: BinaryIO, context: dict[str, Any] | None = None) -> Self:  # type: ignore
+    def _read(
+        cls, stream: BinaryIO, *, context: dict[str, Any] | None = None, endian: Endianness | None = None, **kwargs
```
yunzheng: I vaguely remember our discussion about adding `**kwargs` for future proofing the API (or was it backwards compatibility?).

But I kinda forgot what the use case was. I'm also not sure if this is the best solution anymore, because using `**kwargs` can introduce mistyped keyword arguments, eg: if someone used `endain='little'` it would go unnoticed.

Please help me remember :)

dissect/cstruct/types/base.py
```diff
@@ -71,71 +72,91 @@ def __default__(cls) -> Self:  # type: ignore
...
+    ) -> Self:  # type: ignore
         """Internal function for reading value.
 
         Must be implemented per type.
 
         Args:
             stream: The stream to read from.
             context: Optional reading context.
+            endian: The endianness to use when reading. If not provided, the cstruct's default endianness will be used.
```
yunzheng:
```suggestion
            endian: The endianness to use when reading (little, big, network, <, >, !, @ or =). When omitted, the cstruct instance's own endianness is used.
```

dissect/cstruct/types/base.py
```diff
@@ -144,50 +165,56 @@ def _read_array(cls, stream: BinaryIO, count: int, context: dict[str, Any] | Non
             stream: The stream to read from.
             count: The amount of values to read.
             context: Optional reading context.
+            endian: The endianness to use when reading. If not provided, the cstruct's default endianness will be used.
```
yunzheng:
```suggestion
            endian: The endianness to use when reading (little, big, network, <, >, !, @ or =). When omitted, the cstruct instance's own endianness is used.
```

dissect/cstruct/types/base.py
```diff
@@ -144,50 +165,56 @@ def _read_array(cls, stream: BinaryIO, count: int, context: dict[str, Any] | Non
...
-    def _write_array(cls, stream: BinaryIO, array: list[Self]) -> int:  # type: ignore
+    def _write_array(cls, stream: BinaryIO, array: list[Self], *, endian: Endianness | None = None, **kwargs) -> int:  # type: ignore
         """Internal function for writing arrays.
 
         Allows type implementations to do optimized writing for their type.
 
         Args:
             stream: The stream to read from.
             array: The array to write.
+            endian: The endianness to use when reading. If not provided, the cstruct's default endianness will be used.
```
yunzheng: Note this is in `_write_array`, so should be "when writing":

```suggestion
            endian: The endianness to use when writing (little, big, network, <, >, !, @ or =). When omitted, the cstruct instance's own endianness is used.
```

dissect/cstruct/types/base.py
```diff
@@ -144,50 +165,56 @@ def _read_array(cls, stream: BinaryIO, count: int, context: dict[str, Any] | Non
...
-    def _write_0(cls, stream: BinaryIO, array: list[Self]) -> int:  # type: ignore
+    def _write_0(cls, stream: BinaryIO, array: list[Self], *, endian: Endianness | None = None, **kwargs) -> int:  # type: ignore
         """Internal function for writing null-terminated arrays.
 
         Allows type implementations to do optimized writing for their type.
 
         Args:
             stream: The stream to read from.
             array: The array to write.
+            endian: The endianness to use when reading. If not provided, the cstruct's default endianness will be used.
```
yunzheng: same here, we are writing.

```suggestion
            endian: The endianness to use when writing (little, big, network, <, >, !, @ or =). When omitted, the cstruct instance's own endianness is used.
```

dissect/cstruct/types/base.py
```diff
@@ -71,71 +72,91 @@ def __default__(cls) -> Self:  # type: ignore
...
+        return cls._read(obj, endian=endian)
 
-    def write(cls, stream: BinaryIO, value: Any) -> int:
+    def write(cls, stream: BinaryIO, value: Any, *, endian: AllowedEndianness | None = None) -> int:
         """Write a value to a writable file-like object.
 
         Args:
             stream: File-like objects that supports writing.
             value: Value to write.
+            endian: The endianness to use when writing. If not provided, the cstruct's default endianness will be used.
```
yunzheng:
```suggestion
            endian: The endianness to use when writing (little, big, network, <, >, !, @ or =). When omitted, the cstruct instance's own endianness is used.
```

dissect/cstruct/types/base.py
```diff
@@ -305,5 +334,24 @@ def _is_eof(stream: BinaryIO) -> bool:
...
+    "little": "<",
+    "big": ">",
+}
+
+
+def normalize_endianness(endian: AllowedEndianness) -> Endianness:
+    """Normalize an endianness string to one of the standard format characters."""
+    if endian not in ENDIANNESS_MAP:
+        raise ValueError(f"Invalid endianness: {endian}")
+    return ENDIANNESS_MAP[endian]
```
yunzheng: Prevent two lookups by using EAFP:

```suggestion
def normalize_endianness(endian: AllowedEndianness) -> Endianness:
    """Normalize an endianness string to one of the standard format characters."""
    try:
        return ENDIANNESS_MAP[endian]
    except KeyError:
        raise ValueError(f"Invalid endianness: {endian!r}") from None
```

Not sure if we also want to print the available endian options in the ValueError. A type checker should already know, so maybe not needed.

yunzheng (changes requested):
Some docstring remarks, and a question about **kwargs.

yunzheng:
> Some docstring remarks, and a question about **kwargs.

Oh oops, you added it in the PR message:

> I also intentionally added a **kwargs to the function signature to allow for future expansion without breaking backwards compatibility again.

anyway, my comment about the risks of using **kwargs still stand. In which case would it break backwards compatibility?

dissect/cstruct/types/base.py
Schamper: ~It was for future proofing, in case we want to add additional keyword arguments in the future, it would break custom type implementors. I do agree with your point, and on the one hand we've only added one new keyword argument in 8 years time. On the other, if we _do_ want to add more, it'd probably have to be a major version bump each time. Unless you can think of another way.~

I looked a little closer, but I wanted to keep my original message too. The `kwargs` is only present in the `_read` methods, which will only ever be used/implemented by (custom) type implementations. The "end user" of cstruct will always use the non-sunder-methods `read(s)`, `write`, `dumps`, `uint32(..., endian=)`. In fact, for the "call the type directly" method, this is already a risk. But due to how cstruct works, there's not really anything we can change in that. For all other "public" methods though, this is not a risk, because they do not use `kwargs`.

So, in short, the only time someone would be dealing with these `kwargs` is if they're implementing their own custom type. In which case, this would correctly act as "future proofing" it against any new keyword arguments.

Schamper:
Improved the docstrings and replied to your comment 😄.

Miauwkeru:
If there is a case where someone uses `endian` as a name of a struct attribute there is some strange behavior:

```python
c_def = """
struct test {
  char[4] endian;
};
"""
c_test = cstruct().load(c_def)

>>> c_test.test(endian='help')
<test endian='help'>

>>> c_test.test(b'help', endian='@')
<test endian=b'h'>

>>> c_test.test(b'', endian=b'help')
*** ValueError: Invalid endianness: 'help', expected one of <, >, !, @, =, network, little, big
```

Is it an idea to prefix it with `cs` or let the compilations of the structures crash if it contains that reserved keyword?

Schamper:
@Miauwkeru what exactly is the weird behavior you're speaking of? I added a unit test without checking the behavior first, with how I would expect it to react, and it passed on the first try. So I'm not sure in what sense it's behaving strangely.

Perhaps your strange behavior is because you didn't type your field correctly? It should be `char endian[4]`.

Miauwkeru:
> @Miauwkeru what exactly is the weird behavior you're speaking of? I added a unit test without checking the behavior first, with how I would expect it to react, and it passed on the first try. So I'm not sure in what sense it's behaving strangely.
> 
> Perhaps your strange behavior is because you didn't type your field correctly? It should be `char endian[4]`.

The malformed struct was indeed valid, that fixed some things and due to it I found a different bug

There seems to be an issue with a struct only containing `char endian[4]` that will cause a typerror, but not when you provide it with more data:
```python
ipdb> c_test.test(b"help", endian='<')
*** TypeError: __init__() got multiple values for argument 'endian'
ipdb> c_test.test(b"help\0", endian='<')
<test endian=b'help'>
```

Schamper:
I'd call this an invalid invocation and not a bug. That invocation would in my opinion be the equivalent of this Python code:

```python
In [1]: class A:
   ...:     def __init__(self, a):
   ...:         self.a = a
   ...: 

In [2]: A(1, a=2)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[2], line 1
----> 1 A(1, a=2)

TypeError: A.__init__() got multiple values for argument 'a'
```

You hit this micro optimization: https://github.com/fox-it/dissect.cstruct/blob/91f2347ef234a41eb3ea6ae55defb9e80379e12a/dissect/cstruct/types/structure.py#L74-L82

Which treats structure initialization with exactly one char/bytes field of exactly the same size to be a "new structure invocation", instead of a "parse these bytes" invocation.

dissect/cstruct/types/base.py
```diff
@@ -144,50 +170,60 @@ def _read_array(cls, stream: BinaryIO, count: int, context: dict[str, Any] | Non
...
-                result.append(cls._read(stream, context))
+                result.append(cls._read(stream, context=context, endian=endian, **kwargs))
             return result
 
-        return [cls._read(stream, context) for _ in range(count)]
+        return [cls._read(stream, context=context, endian=endian, **kwargs) for _ in range(count)]
 
-    def _read_0(cls, stream: BinaryIO, context: dict[str, Any] | None = None) -> list[Self]:
+    def _read_0(
+        cls, stream: BinaryIO, *, context: dict[str, Any] | None = None, endian: Endianness | None = None, **kwargs
```
Miauwkeru: The only other remark I see is that the classes that implement the base class don't use `endian: Endianness | None = None` but use `endian: Endianness`. Not a big issue ofc, but can lead to some confusion.

Schamper: I think that was actually a mistake. Those should never be `None`. Fixed in the base class.

dissect/cstruct/types/base.py
Schamper: @yunzheng I've let my boi claude do some performance testing, and apparently the adding of `**kwargs` has quite a negative performance impact. It's less bad in the compiled code path (only adding a slight overhead per "nested structure" call), but for the interpreted parsing it's quite a bit slower (apparently 10-15%).

Gewetensvraag: is it worth to "future proof" like this with that cost? This is the first time in however old cstruct is that we actually add an argument here.

yunzheng: That's a nice catch of our boi Claude, TIL.

For 10-15% overhead I do not think it's worth it. The API might break if it's not compatible but then at least you know immediately. Now maybe the passed argument might not even work then as it would just be swallowed up by `**kwargs`.

Schamper: Agreed. I removed it.

Schamper:
Copilot review remarked this comment: https://github.com/fox-it/dissect.cstruct/pull/144#discussion_r3557685721

Which on second thought is actually fair. I think it's nicer and a bit more predicable if the existing instance is modified, instead of a fresh copy created. It'd also be a bit faster.

Miauwkeru (dismissed):

Miauwkeru (dismissed):

yunzheng:
> Copilot review remarked this comment: [#144 (comment)](https://github.com/fox-it/dissect.cstruct/pull/144#discussion_r3557685721)
> 
> Which on second thought is actually fair. I think it's nicer and a bit more predicable if the existing instance is modified, instead of a fresh copy created. It'd also be a bit faster.

i agree on this, and it's a contextmanager so i would also expect this.

yunzheng:
> If there is a case where someone uses `endian` as a name of a struct attribute there is some strange behavior:

i believe this is a "known" issue right? Some keyword arguments can conflict with the structure field names. I'm sure i've encountered it before.

I'm not sure if that is also the case here, but it does look like it from the example. I'm not sure if there is a solid solution for that, other than using positional arguments.

Schamper:
For the record, the following works:
```python
In [12]: cs.load("""
    ...: struct test {
    ...:     char endian;
    ...: };
    ...: """)
Out[12]: <dissect.cstruct.cstruct.cstruct at 0x1076d4230>

In [13]: cs.test(endian=b"\x01")
Out[13]: <test endian=b'\x01'>
```

It's specifically this case that breaks right now (the input bytes matches the length of a structure that consists solely out of a single field type that map to bytes):
```python
In [14]: cs.test(b"\x01", endian="<")
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[14], line 1
----> 1 cs.test(b"\x01", endian="<")

File ~/Shares/src/dissect/dissect.cstruct/dissect/cstruct/types/structure.py:87, in StructureMetaType.__call__(cls, *args, **kwargs)
     78 def __call__(cls, *args, **kwargs) -> Self:  # type: ignore
     79     if (
     80         cls.__fields__
     81         and len(args) == len(cls.__fields__) == 1
   (...)     85     ):
     86         # Shortcut for single char/bytes type
---> 87         return type.__call__(cls, *args, **kwargs)
     88     if not args and not kwargs:
     89         obj = type.__call__(cls)

TypeError: __init__() got multiple values for argument 'endian'
```

Removing the following "micro optimization" resolves this:
```diff
diff --git i/dissect/cstruct/types/structure.py w/dissect/cstruct/types/structure.py
index 1ba1a88..e8b1342 100644
--- i/dissect/cstruct/types/structure.py
+++ w/dissect/cstruct/types/structure.py
@@ -76,15 +76,6 @@ class StructureMetaType(MetaType):
         return super().__new__(metacls, name, bases, classdict)
 
     def __call__(cls, *args, **kwargs) -> Self:  # type: ignore
-        if (
-            cls.__fields__
-            and len(args) == len(cls.__fields__) == 1
-            and isinstance(args[0], bytes)
-            and issubclass(cls.__fields__[0].type, bytes)
-            and len(args[0]) == cls.__fields__[0].type.size
-        ):
-            # Shortcut for single char/bytes type
-            return type.__call__(cls, *args, **kwargs)
         if not args and not kwargs:
             obj = type.__call__(cls)
             object.__setattr__(obj, "__dynamic_sizes__", {})
```

After:
```python
In [5]: cs.test(b"\x01", endian="<")
Out[5]: <test endian=b'\x01'>
```

I'm fine with removing this micro-optimization too, as it probably doesn't gain us that much anyway.

yunzheng:
what about this:

```python

cs = cstruct("""
struct test1 {
    uint16 foo;
    uint16 bar;
    uint16 endian;
};

struct test2 {
    uint16 endian;
};
""")
```

and then:

```python
$ uv run python -i t.py

>>> print(cs.test1(0x1337, endian=">"))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/workspaces/dissect.cstruct/dissect/cstruct/types/structure.py", line 508, in __repr__
    value = hex(value)
            ^^^^^^^^^^
TypeError: 'str' object cannot be interpreted as an integer

>>> print(cs.test2(0x1337, endian=">"))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/workspaces/dissect.cstruct/dissect/cstruct/types/structure.py", line 93, in __call__
    return super().__call__(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/dissect.cstruct/dissect/cstruct/types/base.py", line 54, in __call__
    return type.__call__(cls, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: __init__() got multiple values for argument 'endian'
```

UPDATE:

ok talked with Schamper, i'm a dummy and don't mind the usecase above. The `endian=` keyword is mainly for initializing structs via bytes/file-like object. Not when using it like a dataclass or normal class.

dissect/cstruct/cstruct.py
```diff
@@ -279,6 +286,25 @@ def add_custom_type(
...
+                if not hasattr(type_to_check, method):
+                    continue
+
+                signature = inspect.signature(getattr(type_to_check, method))
+
+                # Only warn if the method doesn't accept an endian parameter
+                if "endian" not in signature.parameters:
+                    warnings.warn(
+                        f"Custom type {type_name} is missing the 'endian' keyword-only parameter in its {method} method. "  # noqa: E501
+                        "Please refer to the changelog of dissect.cstruct 5.0 for more information.",
```
yunzheng, line 304: don't forget this when creating the changelog. Or do we already keep track of a changelog in the docs?

yunzheng (approved):
LGTM

dissect/cstruct/cstruct.py
Schamper: No not yet, I plan to make a big changelog when we start preparing to release 5.0.

Issue #142, opened 2026-02-03
Add `sizeof` utility function


Schamper:
Could be useful shorthand.

yunzheng:
would be indeed nice, currently you can get the size of a struct by accessing `.size` attribute of the class or an instance. But if you define this as a field in the struct it will break:


```python
from dissect.cstruct import cstruct

example = cstruct("""
struct Example {
    uint32 temp;
    char buffer[1024];
};

struct ExampleWithSize {
    uint32 size;
    char buffer[1024];
};
""")

# works via class
print(f"{example.ExampleWithSize.size=}")
print(f"{example.Example.size=}")

# This still works via instance because size is not defined in the struct.
ex = example.Example()
print(f"{ex.size=}")

# breaks because a field called "size" is defined in the struct.
ex2 = example.ExampleWithSize()
print(f"{ex2.size=}")

# len() does work correctly.
print(f"{len(ex)=}")
print(f"{len(ex2)=}")
```

Outputs:

```
example.ExampleWithSize.size=1028
example.Example.size=1028
ex.size=1028
ex2.size=0
len(ex)=1028
len(ex2)=1028
```

So having a `.size` field available on the instance to get the size is actually confusing. Either the size  attribute should not be exposed and `len()` or `sizeof()` should be the only preferred way.

Note that there are some other predefined fields as well which could interfere in the same way if you would use these names in your struct: https://github.com/fox-it/dissect.cstruct/blob/194b1b556c1474b501d9e120d4750d404786056b/dissect/cstruct/types/base.py#L21-L35

Schamper:
> Note that there are some other predefined fields as well which could interfere in the same way if you would use these names in your struct:

These are meant to be only used on the class objects, not the instances. But unfortunately Python's type model doesn't really allow us to restrict access to that (in the same way that the following is technically a class variable, but still accessible on an instance):

```python
class Foo:
    bar = 1337

foo = Foo()
print(foo.bar)
print(foo.__dict__)
```

Perhaps better documentation could already help. `len()` is already the preferred way to get a structure size, the existence of `sizeof()` could definitely help too.

yunzheng:
What if we use name mangling:

```python
class MyClass:
    __class_only = "I'm meant for the class"
    
    @classmethod
    def get_class_var(cls):
        return cls.__class_only

# Access via class
print(MyClass.get_class_var())  # Works

# Access via instance is mangled
obj = MyClass()
[...]

Issue #138, opened 2025-08-01
add support for const pointer array ```const char* arr[]```


andkae:
Hello Together,

the ```dissect.cstruct``` is a real big deal to parse c files. For ```const *char``` i encounter the error, that ```const``` is unsupported. Following example:

```py
from dissect.cstruct import cstruct
cdef = """
const char* pArrChr[] = {
    "myStr1",   /**< comment */
    "myStr2",   /**< comment */
    "myStr3"
};
"""
cs = cstruct()
cs.load(cdef)
```

Following output:
```
  File "/home/ubuntu/Desktop/dissect.cstruct/dissect/cstruct/cstruct.py", line 268, in load
    TokenParser(self, \*\*kwargs).parse(definition)
  File "/home/ubuntu/Desktop/dissect.cstruct/dissect/cstruct/parser.py", line 470, in parse
    raise ParserError(f"line {self._lineno(token)}: unexpected token {token!r}")
dissect.cstruct.exceptions.ParserError: line 2: unexpected token <Token.IDENTIFIER value='const'>
```

Thanks again for your support.

BR,
Andreas

Schamper:
Hey Andreas, while it would be really nice to support this, unfortunately the current parser is a little too inflexible to easily add support for this syntax. But I'll leave the ticket open for future reference!

Issue #86, opened 2024-04-29
Ignore casts when evaluating #defines


d3dave:
For example:

```c
#define A ((unsigned long) 1)
#define B ((unsigned long) 2)
#define C (A | B)
```

causes A, B and C to have string values like `"(A | B)"` instead of the numeric values.

Schamper:
This looks very fixable. Probably requires dealing with casts in the expression parser though.

@sMezaOrellana can you perhaps recommend the best way to add this to your parser?

sMezaOrellana:
I will try to have a proper look in the coming days. 
In order implement or deal with type-casting it will require changing the evaluate function.

```python
    def evaluate(self, context: Optional[dict[str, int]] = None) -> int:
        """Evaluates an expression using a Shunting-Yard implementation."""

        ...
        while i < len(tmp_expression):
            current_token = tmp_expression[i]
            ...
            elif current_token == "sizeof":
                if len(tmp_expression) < i + 3 or (tmp_expression[i + 1] != "(" or tmp_expression[i + 3] != ")"):
                    raise ExpressionParserError("Invalid sizeof operation")
                self.queue.append(len(self.cstruct.resolve(tmp_expression[i + 2])))
                i += 3
            ...
            elif current_token == "(":
                if i > 0:
                    previous_token = tmp_expression[i - 1]
                    if self.is_number(previous_token):
                        raise ExpressionParserError(
                            f"Parser expected sizeof or an arethmethic operator instead got: '{previous_token}'"
                        )

                self.stack.append(current_token)
            elif current_token == ")":
                if i > 0:
                    previous_token = tmp_expression[i - 1]
                    if previous_token == "(":
                        raise ExpressionParserError(
                            f"Parser expected an expression, instead received empty parenthesis. Index: {i}"
                        )

        ...
        while len(self.stack) != 0:
            if self.stack[-1] == "(":
                raise ExpressionParserError("Invalid expression")

            self.evaluate_exp()

        return self.queue[0]
```
Take for example `((unsigned long) 10)`  this is something like an expression with the left-hand side being `(unsigned long)` and the right-hand side being `10`, there is no explicit operator. 

This could simply evaluate to 10. In practise a simple solution would be to change the case where `elif current_token == "(":` and `elif current_token == ")":`. If whatever is between `(` and `)` is a casting expression just skip over it. It's not very pretty but it would work. In order to actually implement casting we would need to extend the parser, to support evaluating the casting expression.

[...]

PR #146, opened 2026-03-03
Rewrite lexer and parser

### LINKED ISSUE #85: Incorrect parsing of multiple declarators in a single declaration
The definition `struct X { int x : 16, y : 16; }` gets parsed incorrectly into the following tokens:

```python
[<Token.STRUCT value='struct'>,
 <Token.IDENTIFIER value='X'>,
 <Token.BLOCK value='{'>,
 <Token.IDENTIFIER value='int'>,
 <Token.IDENTIFIER value='x'>,
 <Token.NAME value='y : 16'>,
 <Token.EOL value=';'>,
 <Token.BLOCK value='}'>]
```

As you can see, `x : 16` is mistakenly truncated and treated as an identifier instead of a name like `y : 16`.

### LINKED ISSUE #149: ValueError: Cannot use capturing groups in re.Scanner on Python 3.15
When running the tests on Python 3.15 it breaks on:

```
tests/test_align.py:23:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
dissect/cstruct/cstruct.py:267: in load
    TokenParser(self, \*\*kwargs).parse(definition)
dissect/cstruct/parser.py:436: in parse
    scanner = re.Scanner(self.TOK.tokens)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <re.Scanner object at 0x10b73ef90>
lexicon = [('#
[(?P<values>[^
]]+)
](?=
s\*)', <function TokenCollection.add.<locals>.<lambda> at 0x10b9e33d0>), ('#define
s...locals>.<lambda> at 0x10b9e3690>), ('#else
s\*', <function TokenCollection.add.<locals>.<lambda> at 0x10b9e3740>), ...]
flags = 0

    def __init__(self, lexicon, flags=0):
        from ._constants import BRANCH, SUBPATTERN
        if isinstance(flags, RegexFlag):
            flags = flags.value
        self.lexicon = lexicon
1. combine phrases into a compound pattern
           p = []
           s = _parser.State()
           s.flags = flags
           for phrase, action in lexicon:
               sub_pattern = _parser.parse(phrase, flags)
               if sub_pattern.state.groups != 1:
   >               raise ValueError("Cannot use capturing groups in re.Scanner")
   E               ValueError: Cannot use capturing groups in re.Scanner

../../.local/share/uv/python/cpython-3.15.0a2-macos-aarch64-none/lib/python3.15/re/__init__.py:402: ValueError
```

How to reproduce:

```
uv run -p 3.15 pytest -x
```

Exact Python version is: Python 3.15.0a2

Schamper:
Closes #85, partially #142, and will make #86 and #138 a lot easier to implement. Fixes #149.

This PR will (finally) replace the shoddy C syntax parser I originally wrote many moons ago, when I discovered the existence of `re.Scanner` and ran with it. This PR aims to add a somewhat decent lexer and separate parser. I'm still not a compsci 1337coder, so this is just what I came up with (with some help) and definitely not a textbook implementation. All feedback is welcome.

- New lexer
- New C syntax parser that utilizes the new lexer
- Expression parser re-uses the new lexer
- Reworked how `sizeof` works in the expression parser, and added `offsetof`

The new parser has made changing parsing behavior a lot easier. As such, this PR already makes the following changes:

- The new parser is slightly stricter, requiring proper semicolon endings for example. We'll need to fix this in any dissect code that has this.

- An important semantic change is how named nested structures are handled. In my infinite wisdom, I originally figured that named nested structures do not "exist" in the top level scope. That's not true, so now named nested structures get properly registered with the cstruct instance:

```c
struct a {
    struct b {
        ...
    };
};

// Will register both `a` and `b`
```

- Another important change is how we deal with `struct { ... } name;`. We used to parse this first as an anonymous struct, then capture `name` as the structure type name. That's not strictly correct, `name` is a variable of an anonymous unnamed struct, so we now treat it as such. We don't error on this, but rather we silently ignore `name` and skip until we reach a `;`
- `typedef enum ...` is now allowed
- Probably some other things I'm forgetting

This probably warrants a major version bump, so maybe good to pair this with #114, #144 and what we discussed in #142.


Schamper:
@sMezaOrellana would be interested in your thoughts on these changes!

dissect/cstruct/expression.py
```diff
@@ -1,304 +1,203 @@
...
+        while (token := self._current()).type != TokenType.EOF:
+            if token.type == TokenType.NUMBER:
+                self._queue.append(int(self._take().value, 0))
+
+            elif token.type in OPERATORS:
                 while (
-                    len(self.stack) != 0 and self.stack[-1] != "(" and (self.precedence(self.stack[-1], current_token))
+                    len(self._stack) != 0
+                    and self._stack[-1] != TokenType.LPAREN
+                    and precedence(self._stack[-1], token.type)
```
twiggler, line 134: Doesn't make this prefix unary operators left-associative, instead of right-associative?

example: 
`-~1` # Would this work, it must evaluate as `-(~1)?`

Not sure if this is important in practice

dissect/cstruct/parser.py
```diff
@@ -2,712 +2,527 @@
...
+            raise self._error(f"constant {name_token.value!r} not defined", token=name_token)
+
+    def _parse_include(self) -> None:
+        """Parse an include directive and add the included file to the includes list."""
+        self._expect(TokenType.PP_INCLUDE)
+        self.cs.includes.append(self._expect(TokenType.STRING).value)
+
+    def _parse_config_flags(self) -> None:
+        """Parse configuration flags from a directive like ``#[flag1, flag2, ...]``."""
+        self._flags.extend(self._expect(TokenType.PP_FLAGS).value.split(","))
```
twiggler: I don't think  whitespaces are removed? Maybe trim?

twiggler (changes requested):
Architecure looks ok, found 2 possible issues TBC

twiggler:
This PR has been migrated to the dissect monorepo: https://github.com/twiggler/dissect-monorepo-test/pull/5

The original diff and commit history have been preserved on the `migrate/dissect.cstruct/pr-146` branch.

Schamper:
@twiggler I was a bit sad that your message was just this test, as I thought maybe you left more comments 😉.

dissect/cstruct/utils.py
```diff
@@ -360,3 +360,17 @@ def swap64(value: int) -> int:
...
+
+def sizeof(type_: type[BaseType] | BaseType) -> int:
+    """Get the size of a type in bytes."""
+    return len(type_)
+
+
+def offsetof(type_: type[Structure], field: str) -> int:
+    """Get the offset of a field in a structure."""
+    if (field := type_.fields.get(field)) is None:
+        raise ValueError(f"Structure '{type_.__name__}' does not have a field named '{field}'")
```
twiggler: (Found by AI)

Because we overwrite field, won't this always print "does not have a field named None"

dissect/cstruct/parser.py
```diff
@@ -2,712 +2,527 @@
...
-        # Otherwise we're part of a typedef or field definition
-        if tokens.next == self.TOK.EOL:
-            tokens.eol()
+        # If it's still a string, try to evaluate it as an expression in the context of current constants
+        if isinstance(value, str):
+            try:
+                value = Expression(value).evaluate(self.cs)
+            except ExpressionParserError:
+                # If evaluation fails, just keep it as a string (e.g. for macro-like constants)
+                pass
```
twiggler: Is it ok that we propagate LexerError exceptions?

dissect/cstruct/cstruct.py
twiggler: This is no longer true I think?

dissect/cstruct/parser.py
```diff
@@ -2,712 +2,527 @@
...
+    def _at(self, *types: TokenType) -> bool:
+        """Return whether the current token matches any of the given types."""
+        return self._tokens[self._pos].type in types
 
-                if enumtype == "flag":
-                    high_bit = val.bit_length() - 1
-                    nextval = 2 ** (high_bit + 1)
-                else:
-                    nextval = val + 1
+    def _at_value(self, value: str) -> bool:
```
twiggler: I don't think this is used somewhere?

dissect/cstruct/lexer.py
```diff
@@ -0,0 +1,582 @@
...
+
+            elif ch == "<" and self._peek() == "<":
+                self._emit(TokenType.LSHIFT, self._take(2), line, col)
+
+            elif ch == ">" and self._peek() == ">":
+                self._emit(TokenType.RSHIFT, self._take(2), line, col)
+
+            elif ch == "-" and (
+                self._pos == 0
+                or self._tokens[-1].type not in (TokenType.IDENTIFIER, TokenType.NUMBER, TokenType.RPAREN)
```
twiggler: Shouldn't we also add RBRACKET here? as in c[1] - 3

twiggler (changes requested):

twiggler (dismissed):
LGTM 

Since it is an impactful rewrite I asked @Miauwkeru to QA

Miauwkeru:
I did some checking on our other projects to see what goes wrong, and found some regressions regarding to where it crashed unexpectedly.

1. it cannot evaluate a ternary operation, which occurred in `dissect.vmfs` and gives a `LexerError` when trying to evaluate it as it doesn't understand the `?` token. While I am aware that those definitions don't function yet, it probably shouldn't crash when trying to evaluate it

```c
#define func(x) ( x ? 1 : 0 )
```

Without a ternary it makes it into a string:
```c
# define func(x)    ( x == 0)
>>> c_struct.func
' ( x ) ( x = = 0)`
```

2. Another issue I found was when it was trying to create define a flag in dissect.apfs. Here it couldn't evaluate the `.` after the `INODE` in the definition. 

```c
flag INODE {
  ...
};

#define APFS_INODE_PINNED_MASK            INODE.PINNED_TO_MAIN | INODE.PINNED_TO_TIER2
```

I'll look through the code now to see whether I can find some additional issues.

dissect/cstruct/lexer.py
```diff
@@ -0,0 +1,583 @@
...
+
+        if num_newlines := result.count("\n"):
+            self._line += num_newlines
+            self._column = len(result) - result.rfind("\n")
+        else:
+            self._column += num
+
+        return result
+
+    def _expect(self, *chars: str) -> None:
```
Miauwkeru:
```suggestion
    def _expect(self, *chars: str) -> str:
```

tests/test_lexer.py
```diff
@@ -0,0 +1,184 @@
...
+        ("100l", [TokenType.NUMBER], ["100"]),
+        ("100L", [TokenType.NUMBER], ["100"]),
+        ("100ll", [TokenType.NUMBER], ["100"]),
+        ("100ull", [TokenType.NUMBER], ["100"]),
+        ("100ULL", [TokenType.NUMBER], ["100"]),
+        ("0xff", [TokenType.NUMBER], ["0xff"]),
+        ("0XFF", [TokenType.NUMBER], ["0XFF"]),
+        ("0b1010", [TokenType.NUMBER], ["0b1010"]),
+        ("0B1100", [TokenType.NUMBER], ["0B1100"]),
+        ("0755", [TokenType.NUMBER], ["0o755"]),
```
Miauwkeru, line 32: interestingly I tried to use:
- `0888` to test a faulty number and you get a string `0o888` instead of a integer value, while I expected it to fail instead.
- `0o777` resutls in a value of `0 o777` while it should either fail or accept the python style definition?

dissect/cstruct/lexer.py
```diff
@@ -0,0 +1,583 @@
...
+                if condition(ch):
+                    break
+
+            self._pos += 1
+
+        end = self._pos
+        self._pos = start
+        return self._take(end - start)
+
+    def _read_while(self, condition: str | Callable[[str], bool], *, or_eof: bool = True) -> str:
```
Miauwkeru, line 263: What is the difference between `_read_while` and `_read_until` as it looks like they do the same thing

tests/test_lexer.py
```diff
@@ -0,0 +1,184 @@
...
+        ("b'abc'", [TokenType.BYTES], ["b'abc'"]),
+        ('B"hello"', [TokenType.BYTES], ["b'hello'"]),
+        # Identifiers
+        ("b_var", [TokenType.IDENTIFIER], ["b_var"]),
+        ("hello", [TokenType.IDENTIFIER], ["hello"]),
+        ("_my_var", [TokenType.IDENTIFIER], ["_my_var"]),
+        ("UINT32", [TokenType.IDENTIFIER], ["UINT32"]),
+        ("uint32_t", [TokenType.IDENTIFIER], ["uint32_t"]),
+        # Lookup
+        ("$my_lookup = {1: 'a', 2: 'b'}", [TokenType.LOOKUP], ["$my_lookup = {1: 'a', 2: 'b'}"]),
```
Miauwkeru: maybe a stupid question, but how do the lookups work? Cause I am trying to load this specific one inside a cstruct but I get a key error in `_parse_lookup`

Miauwkeru (changes requested):

dissect/cstruct/lexer.py
Schamper: It's the inverse.

tests/test_lexer.py
Schamper: We don't actually ever use it, so I think I'll just remove support for it.

dissect/cstruct/lexer.py
Miauwkeru: Check, I just saw noticed the inversion, kept reading over it for some reason.

I guess you could technically change the condition to always be a callable and only require one of the two functions,  but I assume that would add some unnecessary overhead?

Schamper:
@Miauwkeru fixed.

tests/test_parser.py
```diff
@@ -199,6 +199,58 @@ def test_typedef_pointer(cs: cstruct) -> None:
...
+def test_define_flag_value(cs: cstruct) -> None:
+    cdef = """
+    flag test {
+        VAL1 = 1,
+        VAL2 = 2,
+        VAL3 = 4
+    };
+
+    #define FLAG_VAL1 test.VAL1
+    #define FLAG_VAL3 test.VAL1 | test.VAL2
```
Miauwkeru, line 264: I don't know if this was intended, but feeding it something like:
```
#define MALFORMED_VALUE test.VAL1 | TEST.val1
```
Will return `MALFORMED_VALUE` as `test.VAL1|TEST.val1` which is a fair default. But I do not see a test for this

Miauwkeru:
Only found one more thing that might be an oddity:
```
#define ADCRYPT_MAGIC ADCRYPT\00
```

This now fails to parse due to the `\00` at the end. Which is fixed by using quotes. I don't know if this was an intended change tho

Schamper:
Not necessarily intended. What do you think would be reasonable behavior in this case? (also ping @JSCU-CNI)

Miauwkeru:
> Not necessarily intended. What do you think would be reasonable behavior in this case? (also ping @JSCU-CNI)

I think it makes most sense to go the C route. In the example I gave:
> #define ADCRYPT_MAGIC ADCRYPT\00

C wouldn't compile it as it would expect ADCRYPT to be another definition or something else it can resolve. (Besides not knowing what to do with the `\0`). So I think it would be better if we require strings to be explicitly quoted.

What do you think? @Schamper @JSCU-CNI

Schamper:
I was leaning in that direction too.

JSCU-CNI:
Agreed. Something like `#define ADCRYPT_MAGIC "ADCRYPT\00"` or `#define ADCRYPT_MAGIC b"ADCRYPT\x00"` would make sense.

Schamper:
I changed a bit how `#define` values are handled with https://github.com/fox-it/dissect.cstruct/pull/146/commits/3a37ab0fb5e68af984a110b63aa2499b6213c9fd. Feedback is welcome.

Both ways now actually work.

tests/test_parser.py
```diff
@@ -216,22 +216,36 @@ def test_typedef_enum(cs: cstruct) -> None:
...
-    #define MY_STR "hello"
-    #define MY_BYTES b"world"
-    #define MY_FUNC(x) ( x == 0 )
-    #define MY_TERNARY(x) ( x ? 1 : 0 )
+    #define CONST 42
+    #define EXPR (1 + 2 * 3)
+    #define RAW somevalue
+    #define STR "hello"
+    #define BYTES b"world"
+    #define NULLRAW ADCRYPT\00
```
Miauwkeru, line 224: Don't we want it to be explicitly quoted so that it fails on this kind of definition?

Schamper: My "new approach" (which is basically, don't tokenize anything after `#define NAME`, but just take its raw value until the end of the line) allows this to work again. I think as long as it's unit tested, it should be fine to keep in.

The reason why I slightly prefer this new approach is so that in the parser, we get a more "true" representation of that the define value actually is, including spacing and such. The downside being that the parser now has to deal a little bit with some basic string parsing.

dissect/cstruct/parser.py
```diff
@@ -159,25 +159,38 @@ def _parse_define(self) -> None:
...
+            value = self._take().value
+
+        if value:
+            if value[0] in ('"', "'"):
+                quote = value[0]
+                if value[-1] != quote:
+                    raise self._error("unterminated string literal", token=token)
+
+                # Remove the surrounding and any duplicate quotes
+                value = "".join(ch for ch in value if ch != quote)
```
Miauwkeru: It is pretty much an edge case, but with for example the following define:
`#define test "\'\"a'b\""`

Dissect evaluates it to `"'a'b"` where a c compiler turns it into  `"'\"a'b\""` (checked with godbolt)

Miauwkeru: and another note, adding a `\n` to the string will also create `LexerError: unexpected end of input`

Schamper: This is moreso a limitation of Python than cstruct I think:

```python
In [10]: """abc\"\'\n"""[3]
Out[10]: '"'

In [11]: """abc\"\'\n"""[4]
Out[11]: "'"

In [12]: """abc\"\'\n"""[5]
Out[12]: '\n'
```
There's no way for us to see the escape characters unless you make it a raw string (`r"""`).

The proper way to do it would be to double escape. I've added a test for that and a fix to make sure it works.

Miauwkeru (dismissed):
LGTM, want to release it as 5.0? Then we don't have to update every project immediately.
Otherwise we might need to already have the patch sets ready before it gets merged in.

Schamper:
Yes, see the OP:

> This probably warrants a major version bump, so maybe good to pair this with https://github.com/fox-it/dissect.cstruct/pull/114, https://github.com/fox-it/dissect.cstruct/pull/144 and what we discussed in https://github.com/fox-it/dissect.cstruct/issues/142.

How would we merge this though? Merging now would result in a 4.7.dev and all CI failing.

Miauwkeru:
> How would we merge this though? Merging now would result in a 4.7.dev and all CI failing.

We can yank/delete the `4.7.dev#` it creates on pypi. Ofc not ideal, but it will avoid the CI from falling.
In some cases we might also need to invalidate some caches on projects.

There is also the option to update the publish step to only work when you tag it or disable publishing it in its entirety

Schamper:
Temporarily disabling the publish step is probably the easiest.

Miauwkeru (dismissed):

Schamper:
@Miauwkeru I had to fix the merge with https://github.com/fox-it/dissect.cstruct/pull/140 to work with the removal of lookups so I require a new ack 😄.

Miauwkeru (approved):

JSCU-CNI:
~It seems this merge did cause a pre-release push to PyPi for 4.9 (`dev9`) (https://pypi.org/project/dissect.cstruct/4.8.dev9/#history). Are you planning on yanking that release and bumping it to `5.0.dev1`?~

My bad, that release was for the preceding commit.

Issue #80, opened 2024-03-22
Add color option to `hexdump()` for coloring based on bytes


Schamper:
For example:
-  `00` is darker
- ASCII characters are a certain color (different for lower/upper case/numbers)
- Other color for printable/non-printable
  
Image for inspiration:
![image](https://github.com/fox-it/dissect.cstruct/assets/1254028/4feb883d-41ef-42e0-a688-21db87fb032d)
(From http://thestone.zone/hacking/2019/01/01/hacking-quest-for-glory.html)
  
<sub><sub><sub>the search for chex is real</sub></sub></sub>

yunzheng:
implemented in #155

PR #155, opened 2026-04-23
Add pretty print to hexdump

JSCU-CNI:
This PR aims to provide a stub for pretty colors as described in #80. Currently only null bytes are 'greyed-out' and ASCII printable characters are whitened. 

image

---

Perhaps we can look at some of these resources to determine what other coloring rules we would like to apply here:
* https://simonomi.dev/blog/color-code-your-bytes/
* http://thestone.zone/hacking/2019/01/01/hacking-quest-for-glory.html

Some things that come to mind:
* Should we show difference between certain printable ASCII characters (letters, numbers, capitalization, etc.)?
* Should we color white space characters (`0x09-0x0D`, `0x20`, etc.)?
* What about non-ASCII (`0x80-0xFF`)?

The current implementation does not change the default behavior of `hexdump`, using the `pretty` argument is required.

### DIFF
--- a/dissect/cstruct/utils.py
+++ b/dissect/cstruct/utils.py
@@ -13,6 +13,7 @@ if TYPE_CHECKING:
     from collections.abc import Iterator
     from typing import Literal
 
+# Bold ANSI colors
 COLOR_RED = "\033[1;31m"
 COLOR_GREEN = "\033[1;32m"
 COLOR_YELLOW = "\033[1;33m"
@@ -22,6 +23,10 @@ COLOR_CYAN = "\033[1;36m"
 COLOR_WHITE = "\033[1;37m"
 COLOR_NORMAL = "\033[1;0m"
 
+# Regular ANSI colors
+COLOR_BLACK_REG = "\033[0;30m"
+
+# Background ANSI colors
 COLOR_BG_RED = "\033[1;41m\033[1;37m"
 COLOR_BG_GREEN = "\033[1;42m\033[1;37m"
 COLOR_BG_YELLOW = "\033[1;43m\033[1;37m"
@@ -44,17 +49,41 @@ ENDIANNESS_MAP: dict[str, Literal["big", "little"]] = {
 Palette = list[tuple[int, str]]
 
 
-def _hexdump(data: bytes, palette: Palette | None = None, offset: int = 0, prefix: str = "") -> Iterator[str]:
+def human_colors() -> dict[str, str]:
+    """Generate a dictionary of characters with a human-readable ANSI color they should be in a hexdump."""
+    colors = {
+        "\00": COLOR_BLACK_REG,
+        # ...
+    }
+    # approx. 0x20-0x7E
+    for char in PRINTABLE:
+        colors[char] = COLOR_WHITE
+
+    # ... other coloring rules here ...
+    return colors
+
+
+HUMAN_COLORS = human_colors()
+
+
+def _hexdump(
+    data: bytes, palette: Palette | None = None, offset: int = 0, prefix: str = "", pretty: bool = False
+) -> Iterator[str]:
     """Hexdump some data.
 
     Args:
         data: Bytes to hexdump.
+        palette: Colorize the hexdump using this color pattern.
         offset: Byte offset of the hexdump.
         prefix: Optional prefix.
-        palette: Colorize the hexdump using this color pattern.
+        pretty: Use pretty colors, mutual exclusive with palette.
     """
     if palette:
         palette = palette[::-1]
+
+    # only happy little accidents
+    if pretty and palette:
+        raise ValueError("Cannot use argument 'pretty' in combination with 'palette', please pick one")
 
     remaining = 0
     active = None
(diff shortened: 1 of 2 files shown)


yunzheng:
Nice! I’m a pretty big fan of HexFiend on macOS which also added hex coloring based on hexyl, see this PR:

https://github.com/HexFiend/HexFiend/issues/181

Think if we base the coloring rules and scheme on that it would already be very nice.

JSCU-CNI:
> Think if we base the coloring rules and scheme on that it would already be very nice.

Thanks for the input @yunzheng, I have applied some rules from HexFriend in [9fa9977](https://github.com/fox-it/dissect.cstruct/pull/155/commits/9fa9977c3bb0cb1f53b45464a56466f8895f7cc0). I feel the white color of printable ASCII is somewhat calmer to read than a blue color. What do you think? Perhaps we could make this user customizable in the future, but for simplicity's sake for now I think we should settle on one coloring scheme.

For example, a [small PE file](https://github.com/mathiasbynens/small/) is now rendered like this:

image

yunzheng:
That looks awesome, yes looking at the Hexyl example I think the white is much better on the eyes!

dissect/cstruct/utils.py
```diff
@@ -22,6 +24,13 @@
 COLOR_WHITE = "\033[1;37m"
 COLOR_NORMAL = "\033[1;0m"
 
+# Regular ANSI colors
+COLOR_BLACK_REG = "\033[0;30m"
+COLOR_GREY_REG = "\033[0;90m"
+COLOR_GREEN_REG = "\033[0;32m"
+COLOR_YELLOW_REG = "\033[0;93m"
```
Schamper: Maybe swap the naming so the bold are called `_BOLD` and the regular are just `COLOR_BLACK` etc?

dissect/cstruct/utils.py
```diff
@@ -44,18 +55,55 @@
 Palette = list[tuple[int, str]]
 
 
-def _hexdump(data: bytes, palette: Palette | None = None, offset: int = 0, prefix: str = "") -> Iterator[str]:
+def human_colors() -> dict[str, str]:
```
Schamper:
```suggestion
def _human_colors() -> dict[str, str]:
```

dissect/cstruct/utils.py
```diff
@@ -44,18 +55,55 @@
...
+        colors[char] = COLOR_WHITE
+
+    # Make ASCII whitespace characters green bold (0x9, 0xA, 0xB, 0xC, 0xD, 0x20)
+    for char in ("\t", "\n", "\11", "\12", "\r", "\20"):
+        colors[char] = COLOR_GREEN
+
+    return colors
+
+
+HUMAN_COLORS = human_colors()
```
Schamper:
```suggestion
HUMAN_COLORS = _human_colors()
```

dissect/cstruct/utils.py
```diff
@@ -111,18 +163,31 @@ def _hexdump(data: bytes, palette: Palette | None = None, offset: int = 0, prefi
 
 
 def hexdump(
-    data: bytes, palette: Palette | None = None, offset: int = 0, prefix: str = "", output: str = "print"
+    data: bytes,
+    palette: Palette | None = None,
+    offset: int = 0,
+    prefix: str = "",
+    output: str = "print",
+    pretty: bool | None = None,
```
Schamper, line 179: Can you comment on why `None` is a valid/default value instead of a boolean?

dissect/cstruct/utils.py
```diff
@@ -111,18 +163,31 @@ def _hexdump(data: bytes, palette: Palette | None = None, offset: int = 0, prefi
...
     Args:
         data: Bytes to hexdump.
         palette: Colorize the hexdump using this color pattern.
         offset: Byte offset of the hexdump.
         prefix: Optional prefix.
         output: Output format, can be 'print', 'generator' or 'string'.
+        pretty: Use pretty colors for improved human readability.
     """
-    generator = _hexdump(data, palette, offset, prefix)
+    _pretty = (
```
Schamper: I'm fine with reassigning to `pretty`.

Schamper (changes requested):

dissect/cstruct/utils.py
JSCU-CNI: Because we need to know if `pretty` was explicitly set to `False` by the user or not.

dissect/cstruct/utils.py
```diff
@@ -13,15 +14,29 @@
...
-COLOR_GREEN = "\033[1;32m"
-COLOR_YELLOW = "\033[1;33m"
-COLOR_BLUE = "\033[1;34m"
-COLOR_PURPLE = "\033[1;35m"
-COLOR_CYAN = "\033[1;36m"
-COLOR_WHITE = "\033[1;37m"
-COLOR_NORMAL = "\033[1;0m"
-
+# Regular ANSI colors
+COLOR_RED = ""
```
Schamper: Missing?

dissect/cstruct/utils.py
Schamper: The conditional below is very hard to read, can you maybe split it up over a few `if` branches?

Schamper (changes requested):

Schamper (approved):

PR #156, opened 2026-04-29
Add dict and json output to dumpstruct 

Schamper:
Refactors dumpstruct to add the ability to generate dict and JSON information about a structure. The rendering is also updated to use this dictionary representation.

Also some small maintenance cleanup on the example scripts.

Disclaimer: this change has a fairly high vibe ratio, but it ends up handling quite a few dumpstruct improvements.

### DIFF
--- a/dissect/cstruct/utils.py
+++ b/dissect/cstruct/utils.py
@@ -49,7 +49,6 @@ COLOR_BG_WHITE = "\033[1;47m\033[1;30m"
 
 # Reset ANSI codes
 COLOR_CLEAR = "\033[0m"
-COLOR_CLEAR_BOLD = "\033[1;0m"
 
 PRINTABLE = string.digits + string.ascii_letters + string.punctuation + " "
 
@@ -145,7 +144,7 @@ def _hexdump(
 
                 if active:
                     values += f"{ord(char):02x}"
-                    chars.append(active + print_char + COLOR_CLEAR_BOLD)
+                    chars.append(active + print_char + COLOR_CLEAR)
                 else:
                     if pretty and (color := HUMAN_COLORS.get(char, "")):
                         values += f"{color}{ord(char):02x}{COLOR_CLEAR}"
@@ -159,10 +158,10 @@ def _hexdump(
                     active = None
 
                     if palette is not None:
-                        values += COLOR_CLEAR_BOLD
+                        values += COLOR_CLEAR
 
                 if j == 15 and palette is not None:
-                    values += COLOR_CLEAR_BOLD
+                    values += COLOR_CLEAR
 
             values += " "
             if j == 7:
--- a/dissect/cstruct/utils.py
+++ b/dissect/cstruct/utils.py
@@ -1,14 +1,16 @@
 from __future__ import annotations
 
+import json
 import os
 import pprint
 import string
 import sys
 from enum import Enum
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
+from dissect.cstruct.types.base import BaseArray
 from dissect.cstruct.types.pointer import Pointer
-from dissect.cstruct.types.structure import Structure
+from dissect.cstruct.types.structure import Structure, Union
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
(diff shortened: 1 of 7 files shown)


tests/test_basic.py
```diff
@@ -10,7 +10,7 @@
 from dissect.cstruct.exception import ArraySizeError, ParserError, ResolveError
 from dissect.cstruct.types import BaseType
 
-from .utils import verify_compiled
+from .util import verify_compiled
```
Miauwkeru:
```suggestion
from tests.util import verify_compiled
```

tests/test_bitfield.py
```diff
@@ -5,7 +5,7 @@
 
 import pytest
 
-from .utils import verify_compiled
+from .util import verify_compiled
```
Miauwkeru:
```suggestion
from tests.util import verify_compiled
```

tests/test_types_base.py
```diff
@@ -7,7 +7,7 @@
 from dissect.cstruct.exception import ArraySizeError
 from dissect.cstruct.types.base import BaseArray, BaseType
 
-from .utils import verify_compiled
+from .util import verify_compiled
```
Miauwkeru:
```suggestion
from tests.util import verify_compiled
```

tests/test_types_enum.py
```diff
@@ -6,7 +6,7 @@
 
 import pytest
 
-from .utils import verify_compiled
+from .util import verify_compiled
```
Miauwkeru:
```suggestion
from tests.util import verify_compiled
```

tests/test_types_flag.py
```diff
@@ -8,7 +8,7 @@
 
 from dissect.cstruct.types.enum import PY_311
 
-from .utils import verify_compiled
+from .util import verify_compiled
```
Miauwkeru:
```suggestion
from tests.util import verify_compiled
```

tests/test_types_packed.py
```diff
@@ -4,7 +4,7 @@
 
 import pytest
 
-from .utils import verify_compiled
+from .util import verify_compiled
```
Miauwkeru:
```suggestion
from tests.util import verify_compiled
```

tests/test_types_pointer.py
```diff
@@ -8,7 +8,7 @@
 from dissect.cstruct.exception import NullPointerDereference
 from dissect.cstruct.types.pointer import Pointer
 
-from .utils import verify_compiled
+from .util import verify_compiled
```
Miauwkeru:
```suggestion
from tests.util import verify_compiled
```

tests/test_types_structure.py
```diff
@@ -15,7 +15,7 @@
 from dissect.cstruct.types.pointer import Pointer
 from dissect.cstruct.types.structure import Field, Structure, StructureMetaType
 
-from .utils import verify_compiled
+from .util import verify_compiled
```
Miauwkeru:
```suggestion
from tests.util import verify_compiled
```

tests/test_types_union.py
```diff
@@ -10,7 +10,7 @@
 from dissect.cstruct.types.base import Array, BaseType
 from dissect.cstruct.types.structure import Field, Union, UnionProxy
 
-from .utils import verify_compiled
+from .util import verify_compiled
```
Miauwkeru:
```suggestion
from tests.util import verify_compiled
```

tests/test_types_void.py
```diff
@@ -3,7 +3,7 @@
 import io
 from typing import TYPE_CHECKING
 
-from .utils import verify_compiled
+from .util import verify_compiled
```
Miauwkeru:
```suggestion
from tests.util import verify_compiled
```

Miauwkeru (changes requested):
Some initial things I saw during review, will need to check dump.py a bit more thorough 

And I have a question. When I was testing out dumpstruct I noticed something in the output with the following structure:
```c
struct level_1 {
    struct level_2 {
        struct level_3_1 {
            uint8 r;
            uint8 g;
            uint8 b;
        };
        struct level_3_2 {
            uint8 medicine[3];
        };
    };
};
```

I expect dumpstruct to return
```
struct level_1:
- level_2:
  - level_3_1:
    - r: 0xff
    - g: 0x80
    - b: 0x40
  - level_3_2:
    - medicine[3]: [255, 128, 64]
```

However, what I get is:
```
struct level_1:
- r: 0xff
- g: 0x80
- b: 0x40
- medicine[3]: [255, 128, 64]
```

Is this intended or was this some specific behaviour of c like structs that I was not aware of?

Schamper:
@Miauwkeru none of your nested structures have a field/member name. In C, this would result in a structure with 0 members. You could even argue that cstruct outputs anything at all could be a bug in itself.

```
/tmp ❯ cat x.c
#include <stdio.h>

struct level_1 {
    struct level_2 {
        struct level_3_1 {
            char r;
            char g;
            char b;
        };
        struct level_3_2 {
            char medicine[3];
        };
    };
};

int main() {
    printf("%d\n", sizeof(struct level_1));
};
/tmp ❯ gcc x.c 2>/dev/null && ./a.out
0
```

dissect/cstruct/dump.py
```diff
@@ -0,0 +1,464 @@
...
+def hexdump(
+    data: bytes,
+    *,
+    palette: Palette | None = None,
+    offset: int = 0,
+    prefix: str = "",
+    output: str = "print",
+    pretty: bool | None = None,
+    autoskip: bool = False,
+) -> Iterator[str] | str | None:
```
Miauwkeru, line 197:
```suggestion
AllowedHexdumpOutput: TypeAlias = Literal["print", "generator", "string"]

def hexdump(
    data: bytes,
    *,
    palette: Palette | None = None,
    offset: int = 0,
    prefix: str = "",
    output: AllowedHexdumpOutput = "print",
    pretty: bool | None = None,
    autoskip: bool = False,
) -> Iterator[str] | str | None:
```

Is it an idea to do something like this for the method signature? And then ofc the same for `dumpstruct`

dissect/cstruct/dump.py
```diff
@@ -0,0 +1,464 @@
...
+        counts: list[int | None] = []
+        while hasattr(field_type, "type") and hasattr(field_type, "num_entries"):
+            counts.append(field_type.num_entries)
+            field_type = field_type.type
+        while issubclass(field_type, BaseArray):
+            counts.append(field_type.num_entries)
+            field_type = field_type.type
+
+        if getattr(field_type, "__anonymous__", False):
+            type_name = "union" if issubclass(field_type, Union) else "struct"
```
Miauwkeru, line 309: This might cause some confusion when you are declaring an array of anonymous structs/unions.  every field declared that way would have the type `struct` or `union` but no `fields` section. Where the structure itself is still within `types` by its anonymous name. 
This makes it harder to understand which field belongs to which specific struct.

As an example, this struct:
```c
struct a {
   char magic[4];
};
struct d {
   char footer[4];
};

struct structure {
    struct {
        a yyy;
    } b[4];
    struct {
        d xxx;
    } c[2];
};
```

generates the following overview with `output=json`:
```json
{
    "bytes": "000000000000000000000000000000000000000000000000",
    "root": "structure",
    "types": {
        "structure": {
            "kind": "struct",
            "fields": [
                {
                    "name": "b[4]",
                    "type": "struct"
                },
                {
                    "name": "c[2]",
                    "type": "struct"
                }
            ]
        },
        "a": {
            "kind": "struct",
            "fields": [
                {
                    "name": "magic[4]",
                    "type": "char"
                }
            ]
        },
        "__anonymous_0__": {
            "kind": "struct",
            "fields": [
                {
                    "name": "yyy",
                    "type": "a"
                }
            ]
        },
        "d": {
            "kind": "struct",
            "fields": [
                {
                    "name": "footer[4]",
                    "type": "char"
                }
            ]
        },
        "__anonymous_1__": {
            "kind": "struct",
            "fields": [
                {
                    "name": "xxx",
                    "type": "d"
                }
            ]
        }
    },
    "sizes": {
        "b[4]": 16,
        "c[2]": 8
    },
    "values": {
        "b[4]": "[<__anonymous_0__ yyy=<a magic=b'\\x00\\x00\\x00\\x00'>>,\n <__anonymous_0__ yyy=<a magic=b'\\x00\\x00\\x00\\x00'>>,\n <__anonymous_0__ yyy=<a magic=b'\\x00\\x00\\x00\\x00'>>,\n <__anonymous_0__ yyy=<a magic=b'\\x00\\x00\\x00\\x00'>>]",
        "c[2]": "[<__anonymous_1__ xxx=<d footer=b'\\x00\\x00\\x00\\x00'>>,\n <__anonymous_1__ xxx=<d footer=b'\\x00\\x00\\x00\\x00'>>]"
    }
}
```
Here it isn't clear (except by looking at the values) which types populate `b[4]` and `c[2]` in `structure`

Miauwkeru (changes requested):

PR #158, opened 2026-05-19
Add autoskip option to hexdump and dumpstruct

yunzheng:
This PR adds a `autoskip` argument to the `hexdump()` and `dumpstruct()` functions. When `autoskip=True` it will collapse all consecutive NULL bytes on a hexdump line to a single `*`. This mimics the `--autoskip` feature of `xxd`:

```shell
$ xxd -a /bin/ls | head
00000000: cafe babe 0000 0002 0100 0007 0000 0003  ................
00000010: 0000 4000 0000 bc00 0000 000e 0100 000c  ..@.............
00000020: 8000 0002 0001 0000 0001 5c00 0000 000e  ..........\.....
00000030: 0000 0000 0000 0000 0000 0000 0000 0000  ................
*
00004000: cffa edfe 0700 0001 0300 0000 0200 0000  ................
00004010: 1200 0000 c806 0000 8500 2000 0000 0000  .......... .....
00004020: 1900 0000 4800 0000 5f5f 5041 4745 5a45  ....H...__PAGEZE
00004030: 524f 0000 0000 0000 0000 0000 0000 0000  RO..............
00004040: 0000 0000 0100 0000 0000 0000 0000 0000  ................
```

and with dissect hexdump `autoskip=True`:

image

Note that I have also changed the `hexdump()` function signature to be kwargs only starting from the second argument. I think this is much cleaner and less error prone. I'm doubting if we should do the same for `dumpstruct()`.

### DIFF
--- a/dissect/cstruct/utils.py
+++ b/dissect/cstruct/utils.py
@@ -94,7 +94,13 @@ HUMAN_COLORS = _human_colors()
 
 
 def _hexdump(
-    data: bytes, palette: Palette | None = None, offset: int = 0, prefix: str = "", pretty: bool | None = False
+    data: bytes,
+    *,
+    palette: Palette | None = None,
+    offset: int = 0,
+    prefix: str = "",
+    pretty: bool | None = False,
+    autoskip: bool = False,
 ) -> Iterator[str]:
     """Hexdump some data.
 
@@ -104,6 +110,7 @@ def _hexdump(
         offset: Byte offset of the hexdump.
         prefix: Optional prefix.
         pretty: Use pretty colors, mutual exclusive with palette.
+        autoskip: A single '*' replaces NUL-lines in the output.
     """
     if palette:
         palette = palette[::-1]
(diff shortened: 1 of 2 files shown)


yunzheng:
@codspeedbot explain why this is faster

Schamper (commented):
NUL or NULL?

yunzheng:
> NUL or NULL?

`xxd` calls it NUL in the man page:

```
       -a | -autoskip
              Toggle autoskip: A single '*' replaces NUL-lines.  Default off.
```

Also it's not very clear in the man page, but it always leaves the first NUL(L) line intact.

Schamper (approved):

Schamper:
> I'm doubting if we should do the same for `dumpstruct()`.

Probably, yes. But that has bigger changes with https://github.com/fox-it/dissect.cstruct/pull/156. Perhaps we can do that in that PR.

yunzheng:
> > I'm doubting if we should do the same for `dumpstruct()`.
> 
> Probably, yes. But that has bigger changes with #156. Perhaps we can do that in that PR.

roger

PR #161, opened 2026-07-09
Add declarative syntax

Schamper:
This adds an optional declarative syntax for structure definitions, for when you want to write structures in a more Pythonic syntax.

```python
>>> from dissect.cstruct.declarative import Struct, uint8, uint32, uint64
>>>
>>> class Header(Struct):
...     magic: uint8[4]
...     version: uint32
>>> class File(Struct):
...     header: Header  # structs can reference each other
...     size: uint64
```

Or to bring-your-own-cstruct:
```python
>>> from dissect.cstruct import cstruct
>>> from dissect.cstruct.declarative import Struct
>>>
>>> cs = cstruct(endian=">")
>>>
>>> class Header(Struct, cs=cs):
...     magic: uint8[4]
...     version: uint32
```

The resulting classes are proper subclasses of `Structure` and have all the same behavior as regular cstruct structures.

### DIFF
--- a/dissect/cstruct/declarative.py
+++ b/dissect/cstruct/declarative.py
@@ -1,0 +1,626 @@
+"""Declarative syntax for defining cstruct structures.
+
+This module is an optional convenience layer on top of the existing cstruct. It only offers an alternative,
+Pythonic way to declare structures by subclassing a base class and using type annotations.
+
+Easy usage (no cstruct instance needed):
+    >>> from dissect.cstruct.declarative import Struct, uint8, uint32, uint64
+    >>>
+    >>> class Header(Struct):
+    ...     magic: uint8[4]
+    ...     version: uint32
+    >>> class File(Struct):
+    ...     header: Header  # structs can reference each other
+    ...     size: uint64
+
+Subclassing :class:`Struct` (or :class:`Union`) without any extra arguments gives each structure its own,
+automatically created cstruct instance. A structure that references another (by annotation) is co-located
+onto that structure's instance, so related types share a namespace while unrelated ones stay isolated.
+
+Advanced usage (bind to your own cstruct instance):
+    >>> from dissect.cstruct import cstruct
+    >>> from dissect.cstruct.declarative import Struct
+    >>>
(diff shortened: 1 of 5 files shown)


Miauwkeru (commented):
Some remarks when trying out to use this declarative version.
Currently the `Struct` instance will fail when you try to use `dumps`, `print` or initializing the structure.

Using the `Header` struct from the example, using it like this. it complains about the endianess of the attached cstruct instance, which doesn't exist.

E.g:
- `Header().dumps()`
- `Header(b"heey0102")`

Adding a cstruct instance in the class definition `class Header(..., cs=cs): ...` it cannot find the `__fields__` attribute of the `Struct`.

---
Another thing I noticed is with an empty `Struct` instance, (e.g. `Header()`) where it creates an attribute error with the `magic` and `version` attributes.

Trying to print this `Header` instance will result into another error where it cannot find the `fields` attribute.

dissect/cstruct/declarative.py
```diff
@@ -0,0 +1,626 @@
...
+* Another declarative structure. It is embedded directly by reference, so it may belong to a different
+  cstruct instance (a structure that references another is otherwise co-located onto that instance).
+* A string containing a cstruct type name, e.g. ``"uint8[4]"``.
+* A :func:`field` specification for advanced options such as bit fields or explicit offsets, either
+  directly (``field(uint16, bits=4)``) or, for a type-checker-friendly spelling, as
+  :data:`typing.Annotated` metadata (``Annotated[uint16, field(bits=4)]``).
+
+A structure may reference itself through a pointer (``next: pointer["Node"]`` or ``next: "Node*"``);
+directly embedding a structure in itself is an error. Subclassing a concrete structure extends it:
+the subclass inherits the fields of its parent, followed by its own.
```
Miauwkeru, line 52: shouldn't there also be a note like in https://github.com/fox-it/flow.record/blob/d50435feffc7e13eb3a636e995a27e3e3418eeda/flow/record/declarative.py#L43 for ruff not put the `Annotation` and `field` things underneath `TYPE_CHECKING`?

PR #164, opened 2026-07-13
Small fixes in new parser

Schamper:
Some small fixes I encountered while trying to use the new parser code in `dissect.executable`.

### DIFF
--- a/dissect/cstruct/lexer.py
+++ b/dissect/cstruct/lexer.py
@@ -214,7 +214,11 @@ class Lexer:
         return self._take()
 
     def _error(self, msg: str, *, line: int | None = None) -> LexerError:
-        return LexerError(f"line {line if line is not None else self._line}: {msg}")
+        line = line if line is not None else self._line
+        context = self.data.splitlines()[line - 2 : line + 2]
+        no_len = len(str(line)) + 2
+        content = "\n".join(f"{i + line - 1:>{no_len}}: {l}" for i, l in enumerate(context))
+        return LexerError(f"line {line}: {msg}\n{line}: {content}")
 
     def _emit(self, type: TokenType, value: str, line: int, column: int = 0) -> None:
         """Emit a token with the given type and value at the specified line and column."""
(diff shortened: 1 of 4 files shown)


dissect/cstruct/lexer.py
```diff
@@ -214,7 +214,11 @@ def _expect(self, *chars: str) -> str:
         return self._take()
 
     def _error(self, msg: str, *, line: int | None = None) -> LexerError:
-        return LexerError(f"line {line if line is not None else self._line}: {msg}")
+        line = line if line is not None else self._line
+        context = self.data.splitlines()[line - 2 : line + 2]
+        no_len = len(str(line)) + 2
+        content = "\n".join(f"{i + line - 1:>{no_len}}: {l}" for i, l in enumerate(context))
+        return LexerError(f"line {line}: {msg}\n{line}: {content}")
```
Miauwkeru:
```suggestion
        return LexerError(f"line {line}: {msg}\n{content}")
```

The error output looked a bit messy. So this removes the `419:` from the second line of the error: 
```python
dissect.cstruct.exception.LexerError: line 491: unexpected character '`'
491:   490: "
  491: #define IMAGE_ARCHIVE_END                    b"`
  492: "
  493: #define IMAGE_ARCHIVE_PAD                    b"
```

Miauwkeru (changes requested):

dissect/cstruct/lexer.py
Schamper: Oh yeah that was the first iteration but forgot to remove it after 😄

tests/test_parser.py
```diff
@@ -567,3 +566,27 @@ def test_preprocessor_define_from_enum_in_struct() -> None:
...
+
+    # Error on line 3: shows 1 line before (line 2), error line (3), and 2 lines after (4, 5)
+    # Line 1 is two lines before the error and should be outside the window
+    src = "#define A 1\n#define B 2\n69\n#define C 3\n#define D 4\n#define E 5"
+    with pytest.raises(ParserError, match="line 3:") as exc_info:
+        cs.load(src)
+    message = str(exc_info.value)
+    assert "  1: #define A 1" in message  # 2 line before
+    assert "  2: #define B 2" in message  # 1 line before
+    assert "  3: 69" in message  # error line
```
Miauwkeru: is there a reason not to just check the whole message?
as in:
```
assert message == textwrap.dedent("...")
```
This way no additional weirdness can slip into the error message itself if it gets changed in the future

Miauwkeru (approved):

PR #165, opened 2026-07-15
Move all internal attributes to dunder attributes

### LINKED ISSUE #142: Add `sizeof` utility function

Schamper:
Resolves the comment in https://github.com/fox-it/dissect.cstruct/issues/142. #142 was technically already resolved, but there was another discussion in there. This resolves that.

Tldr, there were a few attributes on all types and structures that could be accessed on the class, but that might be confusing when trying to access them on instances of that class, especially when they conflict with field names in a structure. This change would move all such attributes to dunder attributes.


dissect/cstruct/types/base.py
```diff
@@ -189,6 +190,46 @@ def _write_0(cls, stream: BinaryIO, array: list[Self]) -> int:  # type: ignore
...
+
+    @property
+    def size(cls) -> int | None:
+        """Access the size of this type."""
+        warnings.warn(
+            "The 'size' property is deprecated, use '__size__', len(cls) or sizeof(cls) instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return cls.__size__
```
yunzheng, line 235: for this and the other deprecation properties, how long do we want to keep it around?

yunzheng:
it seems this change has a Performance Regression of -11.35%, any idea why that is?

dissect/cstruct/types/base.py
Schamper: Maybe 5.2? There will also be a number of other "breaking changes" with 5.0, and not all of those have deprecation or fallbacks yet. Will need to add some more.

Schamper:
> it seems this change has a Performance Regression of -11.35%, any idea why that is?

Looks like it resolved itself and was just an issue with a bad luck CI pick.

dissect/cstruct/types/enum.py
```diff
@@ -54,11 +54,11 @@ def __call__(
             raise TypeError("Enum can only be created from int type")
 
         enum_cls = super().__call__(name, *args, **kwargs)
-        enum_cls.cs = cs
+        enum_cls.__cs__ = cs
         enum_cls.type = type_
```
Miauwkeru: Is there a reason this type doesn't use double unders?

dissect/cstruct/compiler.py
```diff
@@ -117,7 +117,7 @@ def generate_source(self) -> str:
         """
 
         if any(field.bits for field in self.fields):
-            preamble += "bit_reader = BitBuffer(stream, endian=endian)\n"
+            preamble += "bit_reader = BitBuffer(stream, endian=cls.__cs__.endian)\n"
```
Miauwkeru: won't it now always use the endianess of the cstruct, regardless of what you give to `_read`?

tests/test_types_custom.py
Miauwkeru: For consistency, what about the dunder for these attributes?

dissect/cstruct/compiler.py
Schamper: Accidental oopsie while rebasing

dissect/cstruct/types/enum.py
Schamper: I forgot about this one, fixed.

dissect/cstruct/cstruct.py
```diff
@@ -206,7 +206,28 @@ def __init__(self, load: str = "", *, endian: AllowedEndianness = "<", pointer:
...
+            | Enum
+            | Flag
+            | Int
+            | Packed[int]
+            | Packed[float]
+            | Pointer
+            | Structure
+            | Union
+            | Void
+            | Wchar
```
Miauwkeru, line 225: wouldn't `type[BaseType]` be enough? as all those types extend that anyways

Schamper: IIRC this gives a little nicer autocompletion hints.

Miauwkeru (approved):

PR #170, opened 2026-07-29
Read a define value untill the end of the line or when it encounters a comment

### LINKED ISSUE #169: Issue when parsing comments after a define
Noticed this when making a patch for `dissect.jffs` that you get unexpected behaviour with a comment after a value with define.

I think this happens at 
https://github.com/fox-it/dissect.cstruct/blob/3032ad71f5e15148222c2faef49e47112a36bf12/dissect/cstruct/lexer.py#L431
as it reads everything to the next `/n`.

Some tests that reproduce it:
```python
def test_define_comment(cs: cstruct) -> None:
    cdef = """
    #define TEST data /* ... */
    """
    cs.load(cdef)
    assert cs.consts["TEST"] == "data"
```

The value of `TEST` here is `data /* ... */`
```python
def test_define_multiline_comment(cs: cstruct) -> None:
    cdef = """
    #define TEST data /* This is the start
                         And here is the end */
    """
    cs.load(cdef)

    assert cs.consts["TEST"] == "data"
```
Where it fails on parsing here due to the unexpected token `And`

Miauwkeru:
- Keeps track of the whitespece between the value and the comment
- Will keep track of a single qouted character

fixes #169


Miauwkeru:
This was the first one that got through all the tests, I will think of a better solution for the quoted string as it will break if you add a single quote somewhere between double quotes

Miauwkeru:
I can replace it with `_read_string()` however, this does require to escape:
https://github.com/fox-it/dissect.cstruct/blob/3032ad71f5e15148222c2faef49e47112a36bf12/tests/test_parser.py#L254

oterwise `_read_string` can't distinguish the quotes from each other. @Schamper is it an idea to always escape the characters? (and document that this is required) If not, I will look for another way

Schamper:
What about multiline defines with quotes at the end of each (or some) line? Note that I'm not sure how that'd work in C, just a thought. Worth a test.

Miauwkeru:
> What about multiline defines with quotes at the end of each (or some) line? Note that I'm not sure how that'd work in C, just a thought. Worth a test.

Do you mean something like this:
```c
#define HELP " \
#define NLD "
```

it won't work without the backslash according to some tests with godbolt

Schamper:
I mean something like this:

```
#define foo bar \  // comment here
            zomg \  // and here
            yolo

**Comment by Miauwkeru** [comment]
> I mean something like this:
> 
> ```
> #define foo bar \  // comment here
>             zomg \  // and here
>             yolo
> ```

it will error in that case in c:
```c
main.c:4:13: error: expected identifier or ‘(’ before string constant
    4 |             "zomg" \  // and here
      |             ^~~~~~
main.c:4:20: error: stray ‘\’ in program
    4 |             "zomg" \  // and here
      |                    ^
main.c:3:19: error: stray ‘\’ in program
    3 | #define foo "bar" \  // comment here
      |                   ^
main.c:8:12: note: in expansion of macro ‘foo’
    8 |     printf(foo);
```

and in cstruct it creates a lexer error for an unexpected `\\` character next to zomg

Schamper:
This is valid C though:

```c
#define foo "bar" /* comment */ \
        "zomg"
```

Might be nice to at least put something like this in the unit test.

Miauwkeru:
> This is valid C though:
> 
> ```c
> #define foo "bar" /* comment */ \
>         "zomg"
> ```
> 
> Might be nice to at least put something like this in the unit test.

I'll add it and see whether I can make that work :)

Miauwkeru:
In `C` there is also the case that you can concatenate two defines if they have the same type. This doesn't work yet as it doesn't resolve yet. I think it might be better to put that functionality in a different pr if we want it.
```python
def test_preprocessor_define_substitute_define(cs: cstruct) -> None:
    """Test whether we can concatinate defines with other strings."""
    c_def = """
    #define DATA1 "world"
    #define CONCAT "hello" \\
            DATA1
    """
    cs.load(c_def)

    assert cs.consts["CONCAT"] == "helloworld"
```

